PR #223, opened 2014-09-11
Html: Only recognize style attributes containing "url(" or "AlphaImageLoader" as inline assets.

papandreou:
I ran into a case where performance degrades a lot because some markup contains tens of thousands of inline style attributes, and each of those is modelled as an inline asset.

This is affecting use cases such as linting CSS as you wouldn't just be able to query for all Css assets. So if we want this, I guess it needs to be put behind a flag. Damn.

Any thoughts?


Munter:
I guess we've slept on this now. I think it makes sense to implement this behavior, possibly with an opt-in flag, which we could activate in the build cases. I'd still like the full CSS to be accessible for programmatic access and refactorings by default.

papandreou:
This is a bad idea, forget it.

PR #226, opened 2014-09-13
Rough implmentation of sub resource ntegrity attribute on ourgoing relat...

### LINKED ISSUE #225: Implement support for sub resource integrity
http://www.w3.org/TR/SRI/

Munter:
...ions from html

Fixes #225


papandreou:
I'd like to dust this off and get it landed.

In addition to what it does now, I think it should support validation of existing `integrity` attributes to encourage a more secure, trust-on-first-use kind of workflow (as mentioned on #225), maybe supplemented by a separate tool that adds `integrity` attributes to the dev html files.

Wrt. the integration into `buildProduction`, which currently doesn't follow relations to assets outside the root, I think we should add the ability to somehow resolve the assetConfig and load the `rawSrc` of external assets that haven't been loaded, but don't put it into the graph, as that would probably break some of the steps in `buildProduction` that makes decisions based on `isLoaded`.

papandreou:
Covered by #504.

PR #229, opened 2014-12-06
Add transform to load and import external CDN assets into the assetgraph root

Munter:
Refs assetgraph/assetgraph-builder#148

@papandreou This is very basic in that it only imports CSS and JavaScript. Might want to add web components as well. The test are running a bit slow, do you have any tricks to stub out the CDN assets?


papandreou:
We can stub the CDN assets out with unexpected-mitm, see the http resolver tests.

Parts of this could also be relevant for SourceMappingURL/SourceURL pointing to an external asset, and for the SRI support where we need to compute hashes of the external assets. We'll need to refine a lot of the queries in buildProduction to do the right thing.

Munter:
Good point. I'll look into that.

How the heck does this PR drop code coverage by 5%?

lib/transforms/importExternalAssets.js
```diff
@@ -1,9 +1,107 @@
...
+                // 'HtmlPictureSourceSrcSet',
+                'HtmlPreconnectLink',
+                'HtmlPrefetchLink',
+                'HtmlPreloadLink',
+                'HtmlPrerenderLink',
+                // 'HtmlRelation',
+                // 'HtmlResourceHint',
+                // 'HtmlScript',
+                'HtmlSearchLink',
+                // 'HtmlServiceWorkerRegistration',
```
papandreou: Should be blacklisted as well.

lib/transforms/importExternalAssets.js
```diff
@@ -1,9 +1,107 @@
...
+                // 'HtmlContentSecurityPolicy',
+                // 'HtmlDataBindAttribute',
+                'HtmlDnsPrefetchLink',
+                'HtmlEdgeSideInclude',
+                'HtmlEdgeSideIncludeSafeComment',
+                // 'HtmlEmbed',
+                // 'HtmlFluidIconLink',
+                'HtmlFrame',
+                'HtmlIFrame',
+                'HtmlIFrameSrcDoc',
```
papandreou: Not necessary to list as it's a "born" inline relation.

Munter:
Updated according to feedback

Issue #240, opened 2015-04-25
Call .toLowerCase() on node.nodeName before comparisons


jscinoz:
Not sure exactly what's causing this (maybe a change in JSDOM?) but I'm noticing that node.nodeName is sometimes all upper-case, whereas all comparisons in assetgraph look for lowercase tag names

This causes things to go horribly wrong in (this is just one specific example I ran into, there are probably others) the href setter in HtmlStyle, as it incorrectly determines the node is not a <link> element, attempts to create a new one, then crashes node entirely when it tries to call replaceChild on a null parent.

```
/home/jcoulter/src/everarchive/node_modules/assetgraph/lib/relations/HtmlStyle.js:34
            this.node.parentNode.replaceChild(link, this.node);
                                ^
TypeError: Cannot read property 'replaceChild' of null
    at HtmlStyle.href (/home/jcoulter/src/everarchive/node_modules/assetgraph/lib/relations/HtmlStyle.js:34:33)
    at HtmlStyle.Relation.refreshHref (/home/jcoulter/src/everarchive/node_modules/assetgraph/lib/relations/Relation.js:121:27)
    at HtmlStyle.<anonymous> (/home/jcoulter/src/everarchive/node_modules/assetgraph/lib/relations/Relation.js:412:26)
    at /home/jcoulter/src/everarchive/node_modules/assetgraph/node_modules/passerror/lib/passError.js:23:33
    at Immediate._onImmediate (/home/jcoulter/src/everarchive/node_modules/assetgraph/lib/index.js:471:17)
    at processImmediate [as _immediateCallback] (timers.js:358:17)

```

jscinoz:
Just to add, according to spec, nodeName is lowerase for XHTML and uppercase for HTML - Perhaps JSDOM made a change to adhere to the spec for this item where it previously did not

jscinoz:
Actually, it looks like toLowerCase is called for most of these comparisons already. It was just missed in the aforementioned getter in HtmlStyle. Having a look if it was missed anywhere else, will send PR shortly.

PR #241, opened 2015-04-25
Ensure all nodeName comparisons are case-insensitive

jscinoz:
Fix for issue #240

### DIFF
--- a/lib/assets/Atom.js
+++ b/lib/assets/Atom.js
@@ -28,10 +28,10 @@ extendWithGettersAndSetters(Atom.prototype, {
             }
 
             if (node.nodeType === 1 &&
-                node.nodeName === 'content' &&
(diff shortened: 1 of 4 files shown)


papandreou:
@jscinoz Please look into the failing tests.

jscinoz:
Sorry for the delay, I've been a bit inactive with code outside of work lately. If I get the chance, I'll review the failing test & address the issue this weekend.

papandreou:
Great, looking forward to it :)

jscinoz:
Seems that making the nodeName comparison in the href setter in the HtmlStyle relation was the cause of the test failure. I can't see just what would cause this to be the case, howerver.

papandreou:
Landed in 847c130. Sorry for the delay :)

PR #246, opened 2015-08-04
Replace the request library with teepee

papandreou:


papandreou:
Oh, the ipv6 thing @Munter ran into with livestyle the other day. Will fix tomorrow.

Munter:
Is it wise to switch to an undocumented library compared to request that most people who know node are at least aware of?

papandreou:
I'm probably not the right one to comment objectively on that, but I do have a strong commitment to keep teepee working and up to date -- and that includes writing documentation once the core feature set is settled. I have spent a few weeks consolidating all HTTP client code in One.com's web apps to use it (including our internal WebDAV client and https://github.com/One-com/beanbag).

My issue with request is that I simply don't trust the code. Also we were stuck on an old version and relied on features that have been removed from the newest version of request (not just deprecated, eg. `onResponse`) so that most of the retrying logic would have to be reworked anyway.

I was in a similar situation with the retrying and agent-handling code in all those other projects, so this is what I went for. AssetGraph might be a slightly different case, but the use of HTTP is very internalized, so I'm not sure that using a familiar and popular library constitutes that much of an advantage in itself.

Munter:
Oh, I wasn't aware that we didn't have a direct upgrade path with request. That does change things a bit I think

papandreou:
The coverage check is punishing me for removing well tested code. It's annoying that it drops below 80%, though.

papandreou:
> Oh, I wasn't aware that we didn't have a direct upgrade path with request. That does change things a bit I think

My biggest worry is still that I don't trust the code. Have you tried skimming through the source?

Munter:
Haven't had time

papandreou:
I've tested this for a bit with `hyperlink` and some internal One.com apps. Everything seems in order.

PR #277, opened 2015-10-10
Updated to unexpected 10 and updated tests and custom assertions

Munter:
@papandreou @sunesimonsen This is my first try with the new assertion syntax. It seems that `'<AssetGraph> to contain [no] (asset|assets) <string|object|number> <number?>'` didn't allow to `expect(graph, 'to have no assets')`, so I had to add a completely duplicate assertion to handle that signature. Is this the right way to go?


sunesimonsen:
We didn't come up with a good syntax for this, the way you currently do it:

```js
expect.addAssertion([
  '<AssetGraph> to contain [no] (asset|assets)',
  '<AssetGraph> to contain [no] (asset|assets) <string|object|number> <number?>'
], function (expect, subject, ...) {
  ...
})
```

sunesimonsen:
Looking at the assertion, I actually think you can get rid of the some the if statements by make to separate assertions.

test/unexpectedAssetGraph.js
```diff
@@ -100,7 +117,23 @@ module.exports = {
...
+                number = queryObj;
+                queryObj = {};
+            }
+            if (this.flags.no) {
+                number = 0;
+            } else if (typeof number === 'undefined') {
+                number = 1;
+            }
+            this.errorMode = 'nested';
+            expect(subject.findRelations(queryObj, this.flags['including unresolved']).length, 'to equal', number);
```
sunesimonsen: You can use to have length

package.json
```diff
@@ -65,10 +65,11 @@
     "sinon": "1.17.1",
     "stylus": "0.52.4",
     "svgo": "0.5.6",
-    "unexpected": "9.6.0",
-    "unexpected-dom": "2.0.1",
-    "unexpected-mitm": "7.7.0",
-    "unexpected-sinon": "6.4.2",
+    "unexpected": "10.0.1",
+    "unexpected-dom": "3.0.0",
+    "unexpected-messy": "5.0.0",
```
papandreou: Can be omitted (although npm 2 lists it after `npm install --save-dev unexpected-mitm`)

test/unexpectedAssetGraph.js
```diff
@@ -71,23 +71,25 @@ module.exports = {
...
-            }
-            if (this.flags.no) {
-                number = 0;
-            } else if (typeof number === 'undefined') {
-                number = 1;
-            }
-            expect(subject.findAssets(value).length, 'to equal', number);
-        });
+        expect.addAssertion([
+            '<AssetGraph> to contain [no] (asset|assets) <string|object|number?>',
```
sunesimonsen: I don't think you need any arguments in the assertion pattern, the rest of the things should be covered by the next line. I think it should just be <AssetGraph> to contain [no] (asset|assets).

PR #279, opened 2015-10-12
JavaScript: Use esprima and escodegen instead of UglifyJS' ast.

papandreou:
Map the `ascii_only` and `indent_level` JavaScript serialization options to escodegen.

Remove the `quoteChar` hack as it was only used to get the UglifyJS code generator to prefer singlequote, which escodegen does by default.

This will require a major version bump, so we might want to land it on a v2 branch until it's stabilized and hopefully bundled with some more improvements. Let's assess later this week.

For testing purposes I've published the current state of the `esprima` branch as assetgraph-esprima@1.23.0-esprima, tracked via the `assetgraph-esprima` branch.


papandreou:
I think this is in a good shape now, so I'd like to land it soon.

@Munter @gustavnikolaj Any thoughts about how or where to land this? I think I vote for putting it on master along with the system.js stuff when it's ready, and anything else we might want in 2.0.0. And if we need to do 1.x.y releases, we can do so from a legacy branch.

Another thing I'd like to get in is https://github.com/assetgraph/assetgraph/issues/296

gustavnikolaj:
Seems like a good plan @papandreou!

Munter:
Sounds like a good plan. In the ideal case we won't need to keep updating the 1.x branch, so let's not deal with a legacy branch of it until needed

papandreou:
Thanks for weighing in. Merging :)

PR #297, opened 2015-10-22
System.js support

papandreou:
This owes a whole lot to @guybedford who wrote half of the code :)

Added partial support for promises to `TransformQueue`.


Munter:
3000+ files? Were the `node_modules` in the test fixtures supposed to be checked in?

papandreou:
Whoa, nice catch. Apparently only a single file in that `node_modules` folder was needed -- fixed that up.

papandreou:
I think this is ready enough to land on master so it can sit there and receive some testing while the remaining 2.0.0 features are settled.

@gustavnikolaj Hit any rough spots during your testing?

Munter:
Sounds good to me :+1:

gustavnikolaj:
@papandreou Not that I can think of. But I haven't used it that intensely since the port I did of the application, when we sat in the room. We can try the state of the two branches before we merge it?

papandreou:
@gustavnikolaj We already landed the esprima branch on master, so I guess it doesn't really make any sense to hold this back unless there's a chance we want to release 2.0.0 without it. The testing might as well happen on master.

gustavnikolaj:
@papandreou You're right!

guybedford:
:+1:

PR #305, opened 2015-10-27
remove engineStrict from package.json

slang800:
Since there's no engine specified, this doesn't do anything. Also, engineStrict was deprecated with npm 3.0.0


papandreou:
Thanks -- seems like an oversight.

PR #312, opened 2015-11-06
ApplicationManifest and all outgoing relations

Munter:
Work in progress...


Munter:
@papandreou The modelling of relations is in place, but this is extremely rudimentary, using `Relation` directly. This means the implementation doesn't have any of the benefits of relation href updates when assets move around, so I guess we need a new type of higher level relation type with this ability.

My current idea would be some sort of `JsonUrl` relation where you also pass a `scope` and a `property` when constructing it. Then `href` getters and setters should be able to update the current instance. Trouble comes when someone overwrites an object with a new instance and the scope reference gets lost. I also thought about using some sort of json query syntax to describe the path to the property that needs an update from the root of the manifest, which could possibly survive more of these manual mutations.

Any ideas?

papandreou:
> so I guess we need a new type of higher level relation type with this ability

Either that or a relation type per unique property that can have a url. Compare: https://github.com/assetgraph/assetgraph/blob/master/lib/relations/SourceMapFile.js

At least that has been my approach so far. Often there are subtle differences that make it hard to use a high level relation.

> Trouble comes when someone overwrites an object with a new instance and the scope reference gets lost. I also thought about using some sort of json query syntax to describe the path to the property that needs an update from the root of the manifest, which could possibly survive more of these manual mutations.

Yeah, a lot of the relations are brittle in that way, especially the ones that only "hold on" to a string value (as opposed to an AST node). It hasn't really turned out to a problem so far -- I guess those manual manipulations aren't really a major use case.

Munter:
Ok. I think this is good to go. Merging

PR #326, opened 2015-11-12
Source map support for JS/CSS, switch from CSSOM to postcss

papandreou:
This seems to be in a state where it's pretty solid and ready to receive feedback.

It would be a bit of work to separate the two things properly as the postcss-related changes ended up requiring a bunch of changes in the source map code I had introduced in order to support JavaScript source maps.

Known issues:

* Inline `<script>` and `<style>` don't get tracked back to the correct line/column in the original HTML. That will require [jsdom.nodeLocation](https://github.com/tmpvar/jsdom/commit/42a1766923acb55905c98a68ad41ac61bc459cef), which landed in jsdom 6.4.0.
* There's a hideous hack at the bottom of lib/assets/Css.js that patches the source-map package that postcss. This is necessary because postcss assumes that the urls are file paths and ends up mangling them. I'll look into fixing it properly in postcss instead.


Munter:
This is awesome!

papandreou:
Other benefits that I forgot to mention:

* postcss supports multiple identically named properties in the same rule, so we no longer need to hack that in (https://github.com/NV/CSSOM/issues/16)
* postcss has a much better minification ([cssnano](https://github.com/ben-eb/cssnano)) and pretty-printing ([perfectionist](https://github.com/ben-eb/perfectionist)) plugins, which I've integrated. Only the "safe" cssnano options are enabled as per http://cssnano.co/options/
* postcss preserves whitespace and comments, which the Css asset now does per default when you've not explicitly called `.minify()` or `.prettyPrint()`
* the source map support also reads source maps referenced via existing `// #sourceMappingUrl=` constructs, including inline ones. Among other things that means it should pick up the correct source locations when running on top of fusile.

papandreou:
Okay, I'll merge this before it gets too unwieldy :)

PR #504, opened 2016-04-26
Support Content Security Policy and add transform for validating and updating it to reflect the state of the graph

papandreou:
Along with https://github.com/assetgraph/assetgraph-builder/tree/feature/contentSecurityPolicy this addresses assetgraph/assetgraph-builder#133


papandreou:
This deviates a bit from [my original proposal](https://github.com/assetgraph/assetgraph-builder/issues/133#issuecomment-51702739) in that it does not attempt to add a CSP to pages that don't have one already. After I read the spec I saw that it would be way too inflexible. Also, nonces are not upgraded to hashes, but hashes will be generated for inline scripts. Lastly, I turned it around a bit so that `buildProduction` does not disable inlining etc. when a policy prohibits it, but changes the policy to reflect what was done instead. That was much easier, and it also less surprising, I think, as you can just use the existing command line options to tell what you want.

papandreou:
One annoying thing about the workflow that this encourages is that if you use inline scripts and stylesheets in dev, you'll need to whitelist `style-src 'unsafe-inline'; script-src 'unsafe-inline'` for your project to work in CSP-compliant browsers. Since the current behavior of `reviewContentSecurityPolicy` is to preserve your existing CSP tokens, `'unsafe-inline'` will thus make it into your production build, which most likely isn't necessary because they will be externalized and bundled in most cases, and the transform adds hashes of the remaining ones.

Maybe we should consider making an exception and remove `'unsafe-inline`' at the end of `reviewContentSecurityPolicy`?

papandreou:
Oh, `'unsafe-inline'` also governs whether inline event handlers and inline styles can be executed, and it appears that those cannot be whitelisted by hashing. You'll just have to avoid using `'unsafe-inline'` in your development version if you don't want it in production.

papandreou:
I implemented upgrading of nonces to hashes and validation of nonce-protected relations. You can use that to avoid `'unsafe-inline'` while still having a working development setup by issuing a bogus nonce:

```html
<meta http-equiv="Content-Security-Policy" content="script-src 'nonce-yeah'">
<script nonce="yeah">
alert("hi");
</script>
```

(note that the nonce must be valid base64)

papandreou:
I think this is ready now, I've successfully integrated it with a non-trivial web application.

There'll be some follow-up work when [CSP3](https://www.w3.org/TR/CSP3/) gets ratified. For one thing, [`'unsafe-hash-attributes'`](https://www.w3.org/TR/CSP3/#unsafe-hash-attributes-usage) will allow hash-signing of inline event handlers and `style` attributes.

papandreou:
@Munter, @gustavnikolaj, or even @mikewest (if I may be so bold) -- any feedback?

lib/assets/ContentSecurityPolicy.js
```diff
@@ -0,0 +1,116 @@
+var util = require('util'),
+    _ = require('lodash'),
+    matchSourceExpression = require('../matchSourceExpression'),
+    extendWithGettersAndSetters = require('../util/extendWithGettersAndSetters'),
+    Text = require('./Text');
+
+function toCamelCase(str) {
```
Munter: Possibly replace with https://www.npmjs.com/package/camelcase

lib/assets/ContentSecurityPolicy.js
```diff
@@ -0,0 +1,116 @@
...
+    return str.replace(/-([a-z])/g, function ($0, ch) {
+        return ch.toUpperCase();
+    });
+}
+
+function fromCamelCase(str) {
```
Munter: Possibly replace with https://www.npmjs.com/package/decamelize

lib/matchSourceExpression.js
```diff
@@ -0,0 +1,180 @@
...
+var defaultPortByScheme = {
+    http: 80,
+    https: 443
+};
+
+function origin(url) {
+    if (/file:\/\//.test(url)) {
+        return 'file:';
+    }
+    var matchOrigin = url.match(/^([a-z][a-z0-9+.-]+):\/\/(\*|(?:\*\.)?[a-z0-9-]+(?:\.[a-z0-9-]+)*)(?::(\d+|\*))?(\/.*)?$/i);
```
Munter: These regexes quickly become unwieldy and hard to test. Could we pull this in from a well tested module?

lib/matchSourceExpression.js
```diff
@@ -0,0 +1,180 @@
...
+
+    // If the source expression matches the grammar for scheme-source (https://www.w3.org/TR/CSP/#scheme_source):
+    if (/^[a-z][a-z0-9+.-]*:$/i.test(sourceExpression)) {
+        // If url’s scheme is an ASCII case-insensitive match for the source expression’s scheme-part, return does match.
+        // Otherwise, return does not match.
+        return urlObj.protocol.toLowerCase() === sourceExpression.toLowerCase();
+    }
+
+    // If the source expression matches the grammar for host-source (https://www.w3.org/TR/CSP/#host_source)
+    var matchHostSource = sourceExpression.match(/^(?:([a-z][a-z0-9+.-]+):\/\/)?(\*|(?:\*\.)?[a-z0-9-]+(?:\.[a-z0-9-]+)*)(?::(\d+|\*))?(\/.*)?$/i);
```
Munter: This regex only has one optional capture group difference to `matchOrigin`, should `matchOrigin` be reused?

lib/matchSourceExpression.js
```diff
@@ -0,0 +1,180 @@
+var urlModule = require('url');
```
Munter: This entire file has no Assetgraph dependencies and the code seems well documented enough to proudly live on its own as a CSP tooling module on npm

lib/transforms/reviewContentSecurityPolicy.js
```diff
@@ -0,0 +1,153 @@
...
+    HtmlAppleTouchStartupImage: 'imgSrc',
+    HtmlVideoPoster: 'imgSrc',
+    HtmlVideo: 'mediaSrc',
+    HtmlAudio: 'mediaSrc',
+    HtmlFrame: 'frameSrc',
+    HtmlIFrame: 'frameSrc',
+    HtmlCacheManifest: 'manifestSrc'
+};
+
+function fromCamelCase(str) {
```
Munter: Possibly use https://www.npmjs.com/package/decamelize

lib/transforms/reviewSubResourceIntegrity.js
```diff
@@ -0,0 +1,53 @@
+var crypto = require('crypto');
+
+module.exports = function (queryObj, options) {
+    options = options || {};
+    return function reviewSubResourceIntegrity(assetGraph) {
+        assetGraph.findAssets(queryObj || { type: 'Html', isInline: false, isFragment: false, isLoaded: true }).forEach(function (htmlAsset) {
+            var changed = false;
+            assetGraph.findRelations({from: htmlAsset, type: ['HtmlStyle', 'HtmlScript'], to: {isInline: false, isLoaded: true}}).forEach(function (relation) {
```
Munter: Only check the loaded ones? There might be a use case for making sure a CDN included asset has the correct identity. Of course this is mostly up the the implementor based on what relationquery is used to populate the graph. Our default queries used in most assetgraph tools usually don't populate external assets, so a security check would be potentially lost here

Munter:
@papandreou Looks good to me. Really awesome that you've picked up those loose ends. CSP and SRI are really awesome functionalities to have support for!

I can't find anything wrong with the functionality as I understand it, only a few improvements, which I've noted. No objections to merging and possibly addressing these after the fact.

I'm not sure I fully understand the workflows that would be recommended with these new capabilities. It might make sense for us to get together and write a blog post or documentation about it at some point

lib/assets/ContentSecurityPolicy.js
papandreou: Ah, I didn't realize that existed. Then again, it took half a minute to type the function, and I'm fairly certain it's bug-free and won't ever have to be updated. Pulling in a dependency feels like overkill.

lib/assets/ContentSecurityPolicy.js
papandreou: Ditto.

lib/matchSourceExpression.js
papandreou: I don't know if such a module exists? I typed them out by hand from the [grammar](https://www.w3.org/TR/CSP/#source-list-syntax). I could split them up into fragments that resemble the productions, then combine them one-validation style.

lib/matchSourceExpression.js
papandreou: Agreed. I thought about doing that at some point. Most of `assets/ContentSecurityPolicy.js`, especially the parsing and normalization code, would also be a good fit for such a module.

lib/transforms/reviewSubResourceIntegrity.js
papandreou: Yes, I agree that it would be very useful to get a warning when an `integrity` attribute that has been manually added to the development build no longer matches. The drawback would be build performance, of course, suggesting that we add a separate tool for it. Also, spinning up the project in the development setting would immediately reveal this situation as the `integrity` attribute would be verified by the browser and the resource would be blocked.

papandreou:
Thanks a lot for reviewing. I starting writing a little bit about the suggested workflow here: https://github.com/assetgraph/assetgraph-builder/tree/feature/contentSecurityPolicy#working-with-a-content-security-policy

It's by no means complete yet, and doesn't cover SRI at all. Would be really great to do a write-up together, perhaps after integrating it with a few more projects.

lib/matchSourceExpression.js
papandreou: Hmm, nice catch. The scheme is mandatory when tokenizing a url, but it should work fine as long as `origin` is only used with correct urls. Let's get back to this once we've decided whether to split the regexps into fragments.

papandreou:
Added some more CSP docs and a short section about SRI.

Munter:
Looks good. We need to work on the SRI story though. The real power unlocks when SRI is applied to third parties to ensure against hacked files. I think we should build a separate tool for adding SRI to external includes. Should be as easy as traversing html --> html relations and finding all non-local scripts and links. Then maybe have a manual verification step with enquirer, so the user explicitly accepts the adding of SRI. I'm not sure how bootstrapper scripts/stylesheets like google analytics or webfonts from external parties will work with this though. Needs further investigation

papandreou:
> I'm not sure how bootstrapper scripts/stylesheets like google analytics or webfonts from external parties will work with this though. Needs further investigation

It'll break as soon as soon as a new version of the external resource is published. That's why I think we'll have to encourage manual whitelisting. We could possibly look at the `Cache-Control: max-age=...` value to learn whether the external resource is likely to ever change, but even that won't be a guarantee.

> I think we should build a separate tool for adding SRI to external includes

Yeah, that would be great. We might face some challenges with not preserving full fidelity when reading, updating, then writing the HTML back to disc. Still, worth the shot.

papandreou:
> >I think we should build a separate tool for adding SRI to external includes
> Yeah, that would be great. We might face some challenges with not preserving full fidelity when reading, updating, then writing the HTML back to disc. Still, worth the shot.

Initial take: https://github.com/assetgraph/assetgraph-builder/commit/d85a02e88c8e743093dc9a472d35d5823765dc4e

Munter:
Looks good.

It might make sense to do a string replacement of just the script/link tags in the original file instead of serialising from assetgraph. This would keep the formatting intact, in case whitespace isn't preserved or there is something jsdom didn't get quite right.

Might also make sense to query the user for each of the possible replacements with inquirer or something similar

papandreou:
> string replacement

... eeek :)

I understand the concern, and we might be able to do a bit better once we get source location support in jsdom (6.something). The translation tools that update the source files suffer from the same problem, but a workflow with `git add -p` works remarkably well, so I'd rather not put string replacements back on the table.

Munter:
Yeah, I would prefer `git add -p` any day. Maybe we should just stick with that until we actually get a user base that doesn't know how to do that :P

papandreou:
Okay, let's roll :)

Released in assetgraph 2.7.0 / assetgraph-builder 4.5.0.

Munter:
Yay!

PR #505, opened 2016-04-27
Add node.js 6 to .travis.yml.

papandreou:
Seems like we're just waiting for https://github.com/bnoordhuis/node-iconv/issues/153 to be fixed, and https://github.com/sass/node-sass/issues/1484 is generating a bit of noise too. Both modules are dev deps with the `^` operator, so let's check back here in a few days and restart the build.


papandreou:
https://github.com/bnoordhuis/node-iconv/issues/153 was fixed, restarted build with my fingers crossed...

PR #522, opened 2016-05-09
Add support for modelling preload relations.

bramstein:
Adds support preload relations so that static asset locations are correctly rewritten. I was tempted to model fonts as proper assets, but I think that goes a little bit beyond the scope of this pull request 😄 .


papandreou:
It just occurred to me that this will make `buildProduction --inline` inline the relation as `data:` url, which does not make any sense for this type of relation. That happens because `HtmlRelation` has an inline method. Took the liberty of landing a fix directly on the branch.

:+1:, feel free to merge.

bramstein:
Thanks, merged it in. What's the policy on doing a release?

papandreou:
There's no policy, just do it :)

After releasing assetgraph you need to update assetgraph-builder to the new version and tag a new release there as well.

PR #524, opened 2016-05-09
Respect source root as prefix for all sources (same like Chrome, etc. do).

swernerx:
Added missing part to prepend the „sourceRoot“ for all files references in „sources“. This seems to be identical to the behavior Chrome shows. „sourceRoot“ can be both: an URL prefix but also just a relative directory reference.


papandreou:
Nice find, thanks a lot for the contribution.

papandreou:
Released in assetgraph 2.5.2 and assetgraph-builder 4.4.2.

PR #527, opened 2016-05-10
Added transform for manipulating the sourceRoot of source maps

swernerx:
This allows for „batch“ updating the source root of source map files. This is super useful when moving sources incl. source maps around using `moveAssetsInOrder` e.g. for flattening/hashing assets. In these cases the sourceRoot has to be removed/resetted to lead to correct resolution of the individual source files.

### DIFF
--- a/lib/transforms/setSourceMapRoot.js
+++ b/lib/transforms/setSourceMapRoot.js
@@ -1,0 +1,18 @@
+var query = require('../query');
+
+module.exports = function (queryObj, root) {
+    var combinedQuery = {type: 'SourceMap'};
+    if (queryObj) {
+        combinedQuery = query.and(queryObj, combinedQuery);
+    }
+
+    return function setSourceMapRoot(assetGraph) {
+        assetGraph.findAssets(combinedQuery).forEach(function (mapFile) {
+            if (root) {
+                mapFile.parseTree.sourceRoot = root;
+            } else {
+                delete mapFile.parseTree.sourceRoot;
+            }
+        });
+    };
+};
--- a/test/transforms/setSourceMapRoot.js
+++ b/test/transforms/setSourceMapRoot.js
@@ -1,0 +1,16 @@
+/*global describe, it*/
+var expect = require('../unexpected-with-plugins'),
+    AssetGraph = require('../../lib');
+
+describe('transforms/setSourceMapRoot', function () {
+    it('should be able to modify source maps', function (done) {
+        new AssetGraph()
+            .loadAssets(new AssetGraph.SourceMap({text: '{"sourceRoot":"rootFolder"}'}))
+            .setSourceMapRoot(null, "otherFolder")
+            .queue(function (assetGraph) {
+                expect(assetGraph, 'to contain asset', 'SourceMap');
+                expect(assetGraph.findAssets({type: 'SourceMap'})[0].parseTree.sourceRoot, 'to equal', 'otherFolder');
+            })
+            .run(done);
+    });
+});
--- a/test/transforms/setSourceMapRoot.js
+++ b/test/transforms/setSourceMapRoot.js
@@ -3,13 +3,24 @@ var expect = require('../unexpected-with-plugins'),
     AssetGraph = require('../../lib');
 
 describe('transforms/setSourceMapRoot', function () {
-    it('should be able to modify source maps', function (done) {
+    it('should be able to modify source root', function (done) {
         new AssetGraph()
             .loadAssets(new AssetGraph.SourceMap({text: '{"sourceRoot":"rootFolder"}'}))
             .setSourceMapRoot(null, "otherFolder")
             .queue(function (assetGraph) {
                 expect(assetGraph, 'to contain asset', 'SourceMap');
                 expect(assetGraph.findAssets({type: 'SourceMap'})[0].parseTree.sourceRoot, 'to equal', 'otherFolder');
+            })
+            .run(done);
+    });
+
+    it('should be able to delete source root', function (done) {
+        new AssetGraph()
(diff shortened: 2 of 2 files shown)


papandreou:
That looks very useful. How are you using (or going to use) it in your workflow? Would you like to integrate it with `buildProduction` or create a standalone CLI tool around it?

swernerx:
That's my gulp task where I use it to copy files from the source folder to "dist":

```js
gulp.task("dist-copy", function(done)
{
  var query = assetgraph.query;
  var includeSources = true;

  new assetgraph({root: "src"})
    .loadAssets("*.html")
    .populate({
      followRelations: {
        hrefType: ["relative", "rootRelative"],
        type: includeSources ?
          function() { return true } :
          query.not(["CssSourceMappingUrl", "JavaScriptSourceMappingUrl"])
      }
    })
    .moveAssetsInOrder({
      isLoaded: true,
      type: query.not([
        "Html"
      ])
    }, function (asset) {
      return "/static/" + asset.md5Hex.substr(0, 8) + asset.extension;
    })
    .setSourceMapRoot(null, null)
    .addCacheManifest()
    .writeAssetsToDisc({}, "dist")
    .run(function (err) {
      if (err) {
        console.error("AssetGraph Error:", err);
      }

      done();
    });
})
```

swernerx:
It looks simple but it was quite a way to have a pretty flat "dist" folder structure where even source maps work correctly.

swernerx:
I'm not sure what `buildProduction` is...

papandreou:
Oh, I see, interesting. Thanks for sharing.

`buildProduction` is the main executable provided by https://github.com/assetgraph/assetgraph-builder -- it's a build system.

papandreou:
It seems like `sourceRoot` is currently only supported when finding outgoing relations: https://github.com/assetgraph/assetgraph/blob/master/lib/assets/SourceMap.js#L73-L87 -- it's not honored when updating the urls found in `sources`. For your transform to work in the general case, that would have to be fixed.

For example:

1) Populate a graph with an existing source map without a sourceRoot, following the `SourceMapSource` relations
2) Introduce a `sourceRoot` using the transform -- this should update the entries in `sources` to be relative to the new source root (unless it's an absolute url)
3) Move one of the source files to a different url -- this should update the corresponding entry in `sources`

I realize that you're mostly using the transform to fix up a wrong value in a generated source map, which is a slightly different use case :)

swernerx:
You are right. The solution is not the full blown real solution for full source map support. It's just some small solution for my issue at hand.

swernerx:
As I currently need that exact functionality can we merge this for now? Because the other task you described is probably something more mid-term we should deal with in a separate issue/task.

papandreou:
Sure, I'll take a look at it tonight. If it's really urgent, you can inline it in your script:

```js
[...]
    .moveAssetsInOrder({
      isLoaded: true,
      type: query.not([
        "Html"
      ])
    }, function (asset) {
      return "/static/" + asset.md5Hex.substr(0, 8) + asset.extension;
    })
    .queue(function setSourceMapRoot(assetGraph) {
        assetGraph.findAssets({type: 'SourceMap'}).forEach(function (mapFile) {
            if (root) {
                mapFile.parseTree.sourceRoot = root;
            } else {
                delete mapFile.parseTree.sourceRoot;
            }
        });
    })
    .addCacheManifest()
[...]
```

papandreou:
I tried implementing my suggestions here: https://github.com/assetgraph/assetgraph/tree/added-transform-for-processing-sourceroot-of-sourcemaps

I included a test for your use case: https://github.com/assetgraph/assetgraph/blob/added-transform-for-processing-sourceroot-of-sourcemaps/test/transforms/setSourceMapRoot.js#L57-L74

Would that work? The second `populate` is not mandatory if you don't need the source assets in the graph -- but `{followRelations: {from: {type: AssetGraph.query.not('SourceMap')}}}` in the first `populate` is mandatory, at least for now.

swernerx:
Looks good I think, but for trying it out in my real project it would be helpful, if it would contain the other open PRs.

papandreou:
Merged the other PR and rebased https://github.com/assetgraph/assetgraph/tree/added-transform-for-processing-sourceroot-of-sourcemaps on master. I hope that's sufficient for you to try it out :)

papandreou:
@swernerx, did you get a chance to test this? I'd be fine with merging this PR with my additional `sourceRoot` changes on top, but would be cool to have the extra assurance that it still works for your use case.

swernerx:
I'll have a look now. Sorry for the delay.

swernerx:
That's perfect. Works absolutely fine for me! Thanks!

swernerx:
Please merge & release... living on this branch for now...

swernerx:
I think we can close this PR, right? Because all of the magic is in your new branch...

papandreou:
Thanks! I merged to master, then cherry-picked my commits on top of that.

papandreou:
Released in 2.8.0. I didn't release a new assetgraph-builder yet -- it seemed like you were just using vanilla assetgraph, right?

swernerx:
Yep, that's right. BTW: do you know why some transforms are just in the builder? e.g. useful things like version addition to html files would be pretty interesting for me too - even without using the builder thing.

papandreou:
It's mostly due to bad historical reasons, and it has been bothering me for a while, too. In 2.9.0 I've moved a bunch of transforms into assetgraph itself:

 * gzip
 * addDataVersionAttributeToHtmlElement
 * autoprefixer
 * setAsyncOrDeferOnHtmlScripts
 * duplicateFavicon
 * stripDebug
 * removeDuplicateHtmlStyles

swernerx:
Wow, cool! Thanks a lot 👍

PR #529, opened 2016-05-11
Fix circular dependency with source maps when using moveAssetsInOrder transform

swernerx:
With this fixed we are able to move everything, including source files references in source maps to a destination directory which gives us the awesomeness of a flat hashed directory not only containing the build files, but also the used source files.

Without this fixed it’s not possible to include source maps at all in the move process at otherwise we get cirular references which breaks the move tool.

### DIFF
--- a/lib/transforms/moveAssetsInOrder.js
+++ b/lib/transforms/moveAssetsInOrder.js
@@ -12,7 +12,10 @@ function findAssetMoveOrderBatches(assetGraph, queryObj) {
 
     assetGraph.findAssets({isInline: false}).forEach(function (asset) {
         if (assetMatcher(asset)) {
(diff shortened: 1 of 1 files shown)


papandreou:
:+1:

I ran into this problem myself when I added support for source maps. Ended up deciding that `buildProduction` didn't need to follow `SourceMapSource` and `SourceMapFile` relations, which avoided the problem there.

It makes perfect sense to special case them in `moveAssetsInOrder` as well.

papandreou:
Don't worry about the build failing, it's something completely unrelated.

swernerx:
When it's okay, can we please merge this to master to send out a new release?

papandreou:
Sure :)

papandreou:
Released in 2.6.0.

PR #531, opened 2016-05-14
setHtmlImageDimensions: Set the width and height attributes of <img> pointing at SVGs that have <svg width=... height=...>

papandreou:
Noticed this when I ran Google Page Speed on https://fontfaceobserver.com/ -- it complains that the ``s don't have `width` and `height`. Those SVGs don't have `width` and `height` attributes on the `<svg>` element either, but get styled via CSS. According to https://thatemil.com/blog/2014/04/06/intrinsic-sizing-of-svg-in-responsive-web-design/ we might still be able to work something out based on defaults. Not sure, though, and not included in this PR -- it'll require some testing.

@bramstein, is the source for your site available somewhere so I can play around with this stuff?


papandreou:
This should be pretty low risk as buildProduction is not presently using this transform. Merging.

PR #540, opened 2016-05-24
Using cssnano-paras20xx v3.6.2-patch1 (required for supporting Node v0.10)

paras20xx:
This change fixes the bug mentioned in:
https://github.com/ben-eb/postcss-normalize-url/pull/12


papandreou:
Thanks!

PR #546, opened 2016-06-01
Add error message if dot command is not available

fastner:
If you call `drawGraph()` and no `dot` command is available it is irritating to get some strange spawn error message. With this patch a clear error message `"dot" not found. Please install graphviz.` is written to the console.


papandreou:
Excellent, thanks!

PR #547, opened 2016-06-07
Rework require.js bundling support

papandreou:
This tears out the modelling of `require(...)`, `define(...)` etc. as first class relations and centralizes all the require.js bundling stuff in a single `bundleRequireJs` transform. Also, it simplifies the system.js support by moving all the code for figuring out the active config and adding polyfilling into the `bundleSystemJs` transform.

Requiring css and less is now supported via https://github.com/guybedford/require-css and https://github.com/guybedford/require-less instead of having specific support hacked in.

- [x] Promisify the `bundleRequireJs` transform
- [x] Figure out how to pick up on .html and .ko templates loaded via a require.js plugin instead of utilizing the `inlineKnockoutTemplates` transform.
- [x] Figure out why the removal of the `liftUpJavaScriptRequireJsCommonJsCompatibilityRequire` transform makes the "should handle a test case with an HTML fragment that has bundleable scripts and stylesheets" test fail in buildProduction. Probably a long standing bug that was somehow masked.

Goes along with https://github.com/assetgraph/assetgraph-builder/tree/feature/reworkRequireJs (not complete yet)


papandreou:
I think this is feature complete now, although I haven't tested it in a large project yet. Please review :)

papandreou:
Sister PR in assetgraph-builder: https://github.com/assetgraph/assetgraph-builder/issues/408

papandreou:
// cc @Munter @Munawwar @gustavnikolaj

Munter:
It's quite a big change, a bit tough to review. But I'll see what I can find.

What will be the new story if I want to replace require.js with almond.js?

papandreou:
> It's quite a big change, a bit tough to review. But I'll see what I can find.

I know, and sorry :). Luckily all the changes in `testdata` can mostly be ignored.

> What will be the new story if I want to replace require.js with almond.js?

`data-almond` is still supported, I've just moved the code that does it into the `bundleRequireJs` transform so bundler-specific code isn't scattered all over the place.

Munter:
I've looked through the changes in the non-tests. Haven't gone into much detail, but overall it looks good and I trust our tests. So Thumbs up from me :)

papandreou:
@Munter, thanks! I think it's ready to merge. It'll require a major version bump, of course, and there's some more stuff in that category that I'll like to get done while we're at it. Opened https://github.com/assetgraph/assetgraph/issues/552 so we can discuss the agenda for that :)

papandreou:
I'm going to merge this now. @Munter @Munawwar, let's use master as the "bleeding edge" branch and create a legacy `v2` release branch if we need to get fixes out before v3.

Munter:
👍

PR #563, opened 2016-07-01
Add webpack support

papandreou:
Goes along with https://github.com/assetgraph/assetgraph-builder/pull/417

Not yet tested on real world projects -- please help us out if you have one. It should be "zero conf" if your `webpack.config.js` is at the root of the folder you're building.

All glory to @gustavnikolaj for doing the research and telling me how to do this :)


papandreou:
I'm going to land this now so that it doesn't rot. Although it's not that well-tested yet, it's not very likely to cause harm, either.

Issue #552, opened 2016-06-15
v3 features


papandreou:
Here's the stuff I'd like to land in the next major version:
- [x] Consolidate the require.js support into a single transform that uses the require.js optimizer to do all the work
- [x] Fix the very complex situation with base assets. This will remove a lot of complexity and means that we cannot support `behavior: url(relative/path/to/foo.htc)`, `AlphaImageLoader(src='relative/path/to/foo.bar')` and `GETSTATICURL('relative/path/to/foo.bar')` anymore.
- [x] Web worker support
- [x] Disallow relative hrefs from HTML fragments and get rid of the stupid `fixBaseAssetsOfUnresolvedOutgoingRelationsFromHtmlFragments` and `assumeThatAllHtmlFragmentAssetsWithoutIncomingRelationsAreNotTemplates` transforms.
- [x] Make relations fully managed by their source asset. This could simplify the API and will get rid of the sometimes huge `assetGraph._relations` array. Won't have to break `assetGraph.findRelations`, I think.
- [x] Come up with a replacement for `GETSTATICURL` based on a system.js plugin.
- [x] Webpack support.
- [x] Get rid of all transpilation.
- [x] Remove the support for angular.js templates.
- [x] Model `'...'.toString('url')` as a relation and map `GETSTATICURL('...')` to that, but with a deprecation warning (https://github.com/assetgraph/assetgraph/pull/569).
- [x] Add `crossorigin` getter to relations (https://github.com/assetgraph/assetgraph/issues/578)

This is all leading up to adding support for incremental builds in some shape or form, which will presumably also require a major version bump. That'll take a while to get right, though, so perhaps we should consider that a v4 thing for now.

// cc @Munter @gustavnikolaj @guybedford

gustavnikolaj:
I'd like to nominate `compileJsxToJs` for removal as well. It's in the same boat as angular support, in my opinion, and it would be covered by the webpack/systemjs support for most cases.

Munter:
Extending Gustavs sentiment I'd like to kill any compileXtoY transform. I'd like to leave it up to SystemJS or webpack, or if not, centralize it by using https://github.com/Munter/tolk or Accord directly during the part of Assetgraph that does the file reading, so we never have to deal with non-browser sources inside the graph

papandreou:
[...]

PR #565, opened 2016-07-02
Remove support for transpiling

papandreou:
As discussed on https://github.com/assetgraph/assetgraph/issues/552

Sister PR: https://github.com/assetgraph/assetgraph-builder/pull/419


papandreou:
The coveralls check fails because of the removal of well covered code. Don't worry about it.

Munawwar:
@gustavnikolaj I am working on adding cache to the systemjs plugin. If sessionStorage and indexedDB isn't available then cache will be turned off.

PR #566, opened 2016-07-02
relations/JavaScriptFetch implementation

Munter:
Models JavaScript `fetch()`-calls

@papandreou I'm a bit in doubt on if it was a good idea to also support `this.fetch()` and `self.fetch()` along with `window.fetch()` and `fetch()`. I actually ran into a regression because requirejs has a `this.fetch()` call in its code without any arguments. What's your take on this?

Also, I couldn't quite figure out how to implement an attach method to JavaScriptFetch. Any ideas?


papandreou:
This is for situations where you're `fetch`ing static assets? Or is it mostly intended service worker scenarios?

I'm hoping we can find a plugin-based replacement for `GETSTATICURL` soon -- could that be used instead?

> @papandreou I'm a bit in doubt on if it was a good idea to also support this.fetch() and self.fetch() along with window.fetch() and fetch(). I actually ran into a regression because requirejs has a this.fetch() call in its code without any arguments. What's your take on this?

Yeah, that's not so great, I bet a lot of libraries have `fetch` methods. It's sort of the same as `self.importScripts`, and those angular template registration calls, although collisions with those are probably rare in practice. If we're careful, the only "damage" will be a warning emitted because the `fetch` argument doesn't correspond to a file on disc, but ambiguities are never great.

> Also, I couldn't quite figure out how to implement an attach method to JavaScriptFetch. Any ideas?

You can just leave it out (throw "not supported") unless there's an actual use case. It'd be a little weird, because `fetch(...)` is a `CallExpression` and requires a `.then(...)` to be really useful.

As for `detach`ment, that's also a problem -- you cannot really tear out the `CallExpression` (unless it's wrapped in an `ExpressionStatement`). For instance, it would be very hard to `detach` this `fetch` call:

```js
var promise = fetch('foobar.json');

promise.then(function (result) {
    console.log("yay");
});
```

Munter:
I was only intending it for static assets.Explicitly ignoring and not complaining over the dynamic uses, but still gives us that advantage of not losing track of an asset that's actually statically hard coded.

This will eventually be able to replace `GETSTATICURL` when all old manual XHR code is replaced by `fetch`. That might be a while though, since it's up the the individual developer and them resolving technical dept

And I do see your point about detaching. I'll remove that again.

Munter:
I'll also remove `this.fetch()` and `self.fetch()`

papandreou:
Still a bit on the fence. I have high hopes for handling this through a system.js or webpack plugin. We're exploring a syntax like this: https://github.com/guybedford/asset-plugin-experiment/blob/master/test/asset-variation/main.js

Munter:
Right. There is a bunch of use cases I didn't think about. So it's probably not going to be a replacement for `GETSTATICURL` in other than the simplest use cases.

PR #569, opened 2016-07-05
WIP: Replace GETSTATICURL with strings and a /* @url */ annotation

papandreou:
I went ahead and tried this, but it appears to be blocked by https://github.com/systemjs/builder/issues/329 -- currently the system.js builder strips comments.


papandreou:
Unfortunately things like `'url(' + 'foo.png' /* @url */ + ')'` don't survive uglification because it'll be constant folded to `'url(foo.png)'`

A couple of other suggestions that work without a function being defined and survive uglification:

```js
String('foo.png', 'url');
'foo.png'.toString('url');
```

Any other ideas?

papandreou:
```js
(''.url, 'foo.png');
```

:)

papandreou:
@Munter votes for `'foo.png'.toString('url')`

PR #571, opened 2016-07-08
todo for conditional bundling

guybedford:
Just including some notes for how this could be integrated.


papandreou:
I tried adapting the code to the new proposed API, although I'm not totally sure I got it right.

What would a more elaborate example of `entryPointBundles` be?

```js
{
  'main.js': [
    {
      conditions: { lang: 'en_us' },
      bundle: 'common'
    },
    {
      conditions: { lang: 'da' },
      bundle: 'common'
    }
  ]
}
```

... Or are the `bundle` attributes actually arrays of bundle names, instead of just a single string (ie. does `[...]` mean it's an array)? And would the bundle names be unique, or do they have to be combined with the conditions to be unique? Does the `deferredParent` attribute live in objects that also contain `conditions` and `bundle`?

papandreou:
Replaced by https://github.com/assetgraph/assetgraph/pull/603

PR #574, opened 2016-07-12
WIP: Rework resolveAssetConfig

papandreou:


papandreou:
Seems like this is in a working state now where it manages to simplify two very centrally used functions, including making them sync. What forced them to be async before was the "wildcard relation" support and some of the custom Ext.js-related resolvers that I've kicked out.

To complete the simplification I dropped the "resolver" abstraction altogether, but we can reintroduce that without losing the benefits of this refactoring. I just can't think of a use case right now. What do you think, @Munter?

papandreou:
Remaining work: Promisify transforms.

lib/assets/Asset.js
```diff
@@ -205,12 +205,12 @@ extendWithGettersAndSetters(Asset.prototype, {
      * @param {Function} cb The callback to invoke when the asset is loaded.
      * @api public
      */
-    load: function (cb) {
+    load: function () {
         var that = this;
         if (that.isLoaded) {
-            setImmediate(cb);
+            return Promise.resolve();
```
Munter: Should the promise resolve with the asset itself?

lib/transforms/loadAssets.js
```diff
@@ -27,22 +27,19 @@ module.exports = function () { // ...
...
                 } else {
-                    asset.load(function (err) {
-                        if (err) {
-                            assetGraph.emit('warn', err);
-                            callback();
-                        } else {
-                            callback(null, asset);
-                        }
-                    });
+                    asset.load().then(function () {
```
Munter: If `asset.load` resolves with the asset itself this could be reduced to

```js
asset.load().catch(assetGraph.emit.bind(null, 'warn'));
```

Not entirely sure where to plug in `.nodeify(callback)` in that construction. Just at the end?

lib/transforms/populate.js
```diff
@@ -36,15 +28,7 @@ module.exports = function (options) {
...
-                        assetGraph.ensureAssetConfigHasType(resolvedAssetConfig.to, function (err) {
-                            if (err) {
-                                err.message += '\nIncluding asset [' + resolvedAssetConfig.type + ']: ' + resolvedAssetConfig.from.nonInlineAncestor.url;
-                                err.asset = resolvedAssetConfig.to;
-                                err.relation = resolvedAssetConfig;
-                                assetGraph.emit('warn', err);
-                                return cb();
-                            }
-
+                        assetGraph.ensureAssetConfigHasType(resolvedAssetConfig.to).then(function () {
```
Munter: `ensureAssetConfigHasType` should also resolve with something relevant, like `resolvedAssetConfig`, so we could decouple these functions

papandreou:
It's a good idea to add a meaningful fulfillment value to those functions. I'll try that, and then I'll have a go at also promisifying at least the `loadAssets` and `populate` transforms.

Munter:
I think adding these fulfillment values might enable us to externalize some of these functions so we could even start testing them directly instead of relying on integration testing on the highest level

papandreou:
Landed on master.

PR #582, opened 2016-08-06
Html anchor no blank opener

Munter:
This transform adds rel="noopener" to HTML anchor tags if:
 - The anchor points at a cross origin target
 - The anchor opens in a new window using target="_blank"
 - The anchor does not have rel="opener" set as an attribute

This is a security feature that disallows the target site access to navigate the origin page using `window.opener.location = 'newUrl'`.

It also removes performance problems in some Chrome versions.

See https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/


papandreou:
:+1:

PR #591, opened 2016-08-19
Add preloadAssets transform

Munter:
This transform adds `<link rel="preload">` for relevant assets.

I had to create a new method `HtmlRelation.attachToHead` to be able to have fine grained control over positioning in the DOM. This method lets you define position relative to a dom node instead of an existing relation, which can be hard to trace down to a DOM position.

Before we merge this we need discussion on the following topics
- [ ] Should the transform take a query object and blindly follow it, or should it be highly opinionated about what gets preloaded and in what order?
- [ ] If we are opinionated, which assets are most important to preload?
- [ ] What is the best order of tags in `<head>` and where do the `<link rel="preload">`-tags go?

### DIFF
--- a/lib/assets/Html.js
+++ b/lib/assets/Html.js
@@ -11,6 +11,23 @@ var util = require('util'),
     JavaScript = require('./JavaScript'),
     AssetGraph = require('../');
 
+// Polyfill HTMLElement.matches
+var Element = jsdom.defaultLevel.HTMLElement;
+if (!Element.prototype.matches) {
+    Element.prototype.matches =
+        Element.prototype.matchesSelector ||
+        Element.prototype.mozMatchesSelector ||
+        Element.prototype.msMatchesSelector ||
+        Element.prototype.oMatchesSelector ||
+        Element.prototype.webkitMatchesSelector ||
+        function (s) {
+            var matches = (this.document || this.ownerDocument).querySelectorAll(s),
+                i = matches.length;
+            while (--i >= 0 && matches.item(i) !== this) {}
+            return i > -1;
+        };
+}
+
 function Html(config) {
     if (typeof config.isFragment !== 'undefined') {
         this._isFragment = config.isFragment;
--- a/lib/relations/HtmlPreloadLink.js
+++ b/lib/relations/HtmlPreloadLink.js
@@ -1,8 +1,16 @@
+/**
+ *  Implementation of http://w3c.github.io/preload/#dfn-preload
+ */
+
 var util = require('util'),
     extendWithGettersAndSetters = require('../util/extendWithGettersAndSetters'),
     HtmlRelation = require('./HtmlRelation');
 
 function HtmlPreloadLink(config) {
+    if (!config.to.url) {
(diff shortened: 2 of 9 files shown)


Munter:
My personal intuition tells me most important assets to preload are potentially renderblocking stuff. CSS, non-async javascript (async javascript as well?), webfonts.

Any other critical asset types I'm forgetting? Any non-critical ones that need consideration?

Regarding placement in `<head>` I would assume that the earlier the better, the fewer bytes are transferred before the preloading starts, the better. I need some input from people knoledgeable about browser loading that can tell me if I'm right or wrong here. I would put the `<link rel="preload">` right at the top, possibly after any `<meta charset="...">` tags if that makes any difference

TheLarkInn:
@jantimon implementing the same for html-webpack-plugin seems like the path to go. I'll have to create an issue.

ebidel:
Generally, resources should be preloaded in the same order they'll eventually be requested in. Not doing so can actually delay page load on slower connections b/c you're filling up the pipeline with resources that aren't needed yet.

You want to put preload links high on the page. Sooner the better so the parser finds them right away.

There's some good info here on h2 push and preload: https://twitter.com/ebidel/status/761016996339134464

jantimon:
Well we looked at it and came to the conclusion that there are no real benefits for the html-webpack-plugin because we handle only markup based assets and preload won't help with that:  

https://www.smashingmagazine.com/2016/02/preload-what-is-it-good-for/#loading-of-late-discovered-resources

> While most markup-based resources are discovered fairly early by the browser’s preloader, not all resources are markup-based.

However webfonts and async JavaScript might perform better with preload (depends on the use case)

TheLarkInn:
Nice write-up. Thank you for looking into this!!!

lib/assets/Html.js
```diff
@@ -11,6 +11,23 @@ var util = require('util'),
...
+        Element.prototype.oMatchesSelector ||
+        Element.prototype.webkitMatchesSelector ||
+        function (s) {
+            var matches = (this.document || this.ownerDocument).querySelectorAll(s),
+                i = matches.length;
+            while (--i >= 0 && matches.item(i) !== this) {}
+            return i > -1;
+        };
+}
+
```
papandreou: Good news, https://github.com/assetgraph/assetgraph/pull/592 should save you from polyfilling `matches`.

Munter:
I think this transform might be a good place to populate and preload webfonts from providers we know assetgraph can populate. Google fonts comes to mind. This could shortcut a css roundtrip on a potentially critical asset. Are there other font providers that don't hide their font url behind a layer of tracking javascript?

papandreou:
Do you think it would be possible to do the import/population of fonts in a separate transform that runs before this one?

Munter:
> Do you think it would be possible to do the import/population of fonts in a separate transform that runs before this one?

It would. So maybe this transform should just look for populated webfonts that are not referenced from inline CSS assets in the html. Then whoever controls the graph can populate what's needed. For the google fonts case we probably need to model the character ranges as well if we haven't already. Just to avoid preloading stuff that is unlikely to be needed

PR #592, opened 2016-08-19
Update jsdom to 9.4.2.

papandreou:
I think I got this working acceptably now.

I was eager to start using `jsdom.nodeLocation(node)` to fix generation of source maps of inline scripts and stylesheets, but turns out we need to wait for jsdom to update the parse5 module to 2.x. The parse5 currently used by jsdom only provides start/end index, but we need line/col to be able to be able to make proper source maps. The work is being tracked here: https://github.com/tmpvar/jsdom/pull/1316

### DIFF
--- a/lib/assets/Html.js
+++ b/lib/assets/Html.js
@@ -226,7 +226,9 @@ extendWithGettersAndSetters(Html.prototype, {
                     'getElementById',
                     'getElementsByName'
                 ].forEach(function (methodName) {
-                    this._parseTree[methodName] = document[methodName].bind(document);
+                    if (typeof document[methodName] === 'function') {
+                        this._parseTree[methodName] = document[methodName].bind(document);
+                    }
                 }, this);
             } else {
                 this._parseTree = document;
--- a/lib/assets/Html.js
+++ b/lib/assets/Html.js
@@ -2,8 +2,7 @@
 var util = require('util'),
     _ = require('lodash'),
     esprima = require('esprima'),
-    jsdom = require('jsdom-papandreou'),
-    domtohtml = require('jsdom-papandreou/lib/jsdom/browser/domtohtml'),
+    jsdom = require('jsdom'),
     htmlMinifier = require('html-minifier'),
     extendWithGettersAndSetters = require('../util/extendWithGettersAndSetters'),
     errors = require('../errors'),
@@ -161,7 +160,7 @@ extendWithGettersAndSetters(Html.prototype, {
     get internalText() {
         if (typeof this._text !== 'string') {
             if (this._parseTree) {
-                this._text = domtohtml.domToHtml(this._parseTree, !this.isPretty);
+                this._text = this.isFragment ? this._parseTree.innerHTML : jsdom.serializeDocument(this._parseTree);
             } else {
                 this._text = this._replaceTemplates(this._getTextFromRawSrc());
             }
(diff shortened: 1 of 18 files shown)


papandreou:
https://github.com/assetgraph/assetgraph-builder/commit/0e35f6f72e5b066dbb3921764e1d39a8b4c078bd

papandreou:
@Munter, are you OK with the overall approach?

We seem to lose the ability to use `<esi:include src=... />` and placeholders for translation keys inside the `<title>...</title>` element. This happens because `<title>` cannot contain non-text nodes, and newer jsdom versions enforce this. If this is important it seems like we can restore the previous behavior by adding the child nodes after the document has been created (consistent with behavior in Chrome):

```js
> document = jsdom.jsdom('<!DOCTYPE html><html><head><title>foo<span>bar</span>quux</title></head><body></body></html>')
> document.documentElement.outerHTML
'<html><head><title>foo&lt;span&gt;bar&lt;/span&gt;quux</title></head><body></body></html>'
> document.head.childNodes[0].appendChild(document.createElement('span'))
> document.documentElement.outerHTML
'<html><head><title>foo&lt;span&gt;bar&lt;/span&gt;quux<span></span></title></head><body></body></html>'
```

Also, the new jsdom version no longer exposes the ability to automatically reindent the source when serializing, and `htmlAsset.prettyPrint()` was wired up to do that. We can fix that by implementing the indentation by adjusting the text nodes in the document instead, sort of what `htmlAsset.minifyAsset()` does. I might attempt that.

lib/assets/Html.js
```diff
@@ -214,9 +213,25 @@ extendWithGettersAndSetters(Html.prototype, {
...
+                // Install an assortment of bound methods from the document object on the parse tree:
+                [
+                    'createElement', 'createElementNS',
+                    'createTextNode', 'createComment', 'createDocumentFragment', 'createProcessingInstruction', 'importNode', 'adoptNode',
+                    'createAttribute', 'createAttributeNS',
+                    'createEvent',
+                    'createTreeWalker',
+                    'createNodeIterator',
+                    'getElementById',
+                    'getElementsByName'
```
Munter: `querySelector`, `querySelectorAll` would be really useful here

papandreou: Already present on `document.body` and all other nodes:

```
> require('jsdom').jsdom('<!DOCTYPE html><html><body><div></div></body></html>').body.querySelector('div')
HTMLDivElement {}
```

Munter: How much would break if we do an `Object.keys(this._parseTree)` here?

Munter:
Would it make sense to use a `documentFragment` instead of passing off the `<body>` as the document`?

I'm ok with the small loss of tags in `<title>` and having to figure out pretty printing ourself

lib/assets/Html.js
papandreou: Not too much if we leave out the properties that already exist on `document.body`.

Here's the complete list:

```js
> var doc = require('jsdom').jsdom('<!DOCTYPE html><html><body></body></html>'); for (var p in doc) {if (!(p in doc.body)) {console.log(p, typeof doc[p]);}}
location object
getElementById function
createElement function
createElementNS function
createDocumentFragment function
createTextNode function
createComment function
createProcessingInstruction function
importNode function
adoptNode function
createAttribute function
createAttributeNS function
createEvent function
createTreeWalker function
getElementsByName function
open function
close function
write function
writeln function
implementation object
URL string
documentURI string
origin string
compatMode string
characterSet string
charset string
inputEncoding string
contentType string
doctype object
documentElement object
referrer string
cookie string
lastModified string
readyState string
body object
head object
images object
links object
forms object
currentScript object
defaultView object
activeElement object
onreadystatechange undefined
anchors object
applets object
styleSheets object
visibilityState string
onvisibilitychange undefined
createExpression function
createNSResolver function
evaluate function
createNodeIterator function
```

Some of them don't make a lot of sense to expose (eg. `head`), but I don't know if it would actually harm anything.

papandreou: Implemented in 764d844 (my intuition doesn't fail me, adding `head` actually broke a couple of tests, so I ended up skipping that :)

papandreou:
It would probably work, but I can't find a good way to serialize the document fragment. `jsdom.serializeDocument` only works with documents, and a document fragment doesn't have `innerHTML`. Unless I'm overlooking something in jsdom and/or the DOM spec, we would have to implement something like https://www.w3.org/TR/2009/WD-html5-20090212/serializing-html-fragments.html#html-fragment-serialization-algorithm ourselves :/

Munter:
Ok, better not go down that road. I think this looks good enough to go with it

papandreou:
Thanks. I'll merge it. If we figure out a way to serialize a document fragment, we should be able to introduce it without any breakage later. I might also look into implementing pretty-printing at a later point.

PR #618, opened 2016-09-12
reviewContentSecurityPolicy: Issue absolute links when updating the CSP instead of whitelisting whole host names.

papandreou:
Addresses https://blog.0daylabs.com/2016/09/09/bypassing-csp/

In an app I'm working on, the resulting CSP ends up like this:

```
Content-Security-Policy: default-src 'self';
 script-src 'self' 'sha256-nSGNONqPM3ZufPUUlgPHbZz+04ecLTSSHwDZ/mBGb/s='
   *.google-analytics.com http://cdn.example.com/yay/bundle-4834.1cc9740e94.js;
 style-src 'self' 'unsafe-inline'
   http://cdn.example.com/yay/bundle-4830.8305e75e8c.css; img-src * data:;
 font-src 'self' http://cdn.example.com/yay/OpenSans-Bold-Cyrillic.2713d6aa91.woff2
   http://cdn.example.com/yay/OpenSans-Bold-Cyrillic.3d4530707c.woff
   http://cdn.example.com/yay/OpenSans-Bold-Latin.766fc545e6.woff2
   http://cdn.example.com/yay/OpenSans-Bold-Latin.c681a43c6b.woff
   http://cdn.example.com/yay/OpenSans-Bold-LatinExt.8d0e782b66.woff2
   http://cdn.example.com/yay/OpenSans-Bold-LatinExt.95feb93674.woff
   http://cdn.example.com/yay/OpenSans-BoldItalic-Cyrillic.13d4c28b0d.woff
   http://cdn.example.com/yay/OpenSans-BoldItalic-Cyrillic.7afbe9f843.woff2
   http://cdn.example.com/yay/OpenSans-BoldItalic-Latin.31f14124c4.woff2
   http://cdn.example.com/yay/OpenSans-BoldItalic-Latin.9010f065b6.woff
   http://cdn.example.com/yay/OpenSans-BoldItalic-LatinExt.2e2a1fd9bf.woff
   http://cdn.example.com/yay/OpenSans-BoldItalic-LatinExt.783083a3c4.woff2
   http://cdn.example.com/yay/OpenSans-Cyrillic.85759f5453.woff2
   http://cdn.example.com/yay/OpenSans-Cyrillic.f62f8e01d1.woff
   http://cdn.example.com/yay/OpenSans-Italic-Cyrillic.aa8cecbe1e.woff
   http://cdn.example.com/yay/OpenSans-Italic-Cyrillic.d8098767b8.woff2
   http://cdn.example.com/yay/OpenSans-Italic-Latin.2910f08ca9.woff
   http://cdn.example.com/yay/OpenSans-Italic-Latin.966bd8e51c.woff2   
   http://cdn.example.com/yay/OpenSans-Italic-LatinExt.ccd11a2ef6.woff
   http://cdn.example.com/yay/OpenSans-Italic-LatinExt.e05e5930d0.woff2
   http://cdn.example.com/yay/OpenSans-Latin.4c9a4bc2b3.woff2
   http://cdn.example.com/yay/OpenSans-Latin.90c17a03cb.woff
   http://cdn.example.com/yay/OpenSans-LatinExt.b9e80ad4fc.woff2
   http://cdn.example.com/yay/OpenSans-LatinExt.c043849aad.woff
```

instead of merely:

```
Content-Security-Policy: default-src 'self';
 script-src 'self' 'sha256-nSGNONqPM3ZufPUUlgPHbZz+04ecLTSSHwDZ/mBGb/s='
   *.google-analytics.com cdn.example.com;
 style-src 'self' 'unsafe-inline' cdn.example.com;
 img-src * data:;
 font-src 'self' cdn.example.com
```

... But better safe than sorry, I suppose :)


Munter:
Hopefully font subsetting will let us reduce the amount of font files needed :/

papandreou:
> Hopefully font subsetting will let us reduce the amount of font files needed :/

In this case, I think that'll only make it worse, as it'll introduce even more variants of that font.

Munter:
I thought it would reduce the amount by a factor 3 by pulling relevant glyphs from Latin, LatinExit and Cyrilic into one single font file. Then you would only need regular, bold, italic, bolditalic variants of that subsetted font

papandreou:
In this case the CSS looks already contains 12 `@font-face` declarations:

```css
/* latin */
@font-face {
  font-family: 'Open Sans';
  font-style: normal;
  font-weight: 400;
  src: url(typeface/OpenSans-Latin.woff2) format('woff2'),
       url(typeface/OpenSans-Latin.woff) format('woff');
  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2212, U+2215, U+E0FF, U+EFFD, U+F000;
}
/* latin */
@font-face {
  font-family: 'Open Sans';
  font-style: normal;
  font-weight: 700;
  src: url(typeface/OpenSans-Bold-Latin.woff2) format('woff2'),
       url(typeface/OpenSans-Bold-Latin.woff) format('woff');
  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2212, U+2215, U+E0FF, U+EFFD, U+F000;
}
/* latin */
@font-face {
  font-family: 'Open Sans';
  font-style: italic;
  font-weight: 400;
  src: url(typeface/OpenSans-Italic-Latin.woff2) format('woff2'),
       url(typeface/OpenSans-Italic-Latin.woff) format('woff');
  unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02C6, U+02DA, U+02DC, U+2000-206F, U+2074, U+20AC, U+2212, U+2215, U+E0FF, U+EFFD, U+F000;
}
```

... But of course it might be possible to change that to rely on the automatic or semiautomatic subsetting.

Munter:
https://twitter.com/_munter_/status/775254953728286720

Munter:
From that twitter discussion it looks like it might make sense to differ between the different directives. fonts might not be as vulnerable to the described attack as script

papandreou:
That's a really good point.

papandreou:
Changed the implementation so absolute links are only issued for the `script-src`, `style-src`, `frame-src`, `object-src`, `manifest-src`, and `child-src` directives.

papandreou:
Now that app is down to:

```
Content-Security-Policy: default-src 'self';
 script-src 'self' 'sha256-nSGNONqPM3ZufPUUlgPHbZz+04ecLTSSHwDZ/mBGb/s='
   *.google-analytics.com
   http://cdn.example.com/yay/bundle-25868.1cc9740e94.js;
 style-src 'self' 'unsafe-inline' http://cdn.example.com/yay/bundle-24677.8305e75e8c.css;
 img-src * data:;
 font-src 'self' cdn.example.com
```

Munter:
That sounds like a good compromise. So now we only run into massive policies when people essentially don't bundle, which is rarely the case for csp-aware developers. It's good to get the policy cut down so there aren't to many round trips before any useful resource can start loading in the client

papandreou:
@Munter Okay, so you're :+1: in this state? :)

Munter:
👍

PR #622, opened 2016-09-15
Implement open graph relations support

### LINKED ISSUE #614: Model OpenGraph URL properties
http://ogp.me/

```html
<meta property="og:url" content="http://www.imdb.com/title/tt0117500/" />
<meta property="og:image" content="http://ia.media-imdb.com/images/rock.jpg" />
<meta property="og:audio" content="http://example.com/bond/theme.mp3" />
<meta property="og:video" content="http://example.com/bond/trailer.swf" />
```

Include the structured properties:

```
og:image:url
og:image:secure_url
og:audio:url
og:audio:secure_url
og:video:url
og:video:secure_url
```

Munter:
Closes #614 

Possibly not ready yet!

Issues to be resolved:
- [x] Should we (can we) get the content type from a sibling node of `<meta property="og:image:type" content="image/jpeg" />` ? **Conclusion: Lets rely on the existing type detection**
- [x] Should each type and sub type have their own relation instead of them all sharing the same and relying on `ogProperty` to figure out what property to set? **Conclusion: we don't care for now**

### DIFF
--- a/lib/assets/Html.js
+++ b/lib/assets/Html.js
@@ -582,6 +582,17 @@ extendWithGettersAndSetters(Html.prototype, {
                                 node: node
                             }));
                         }
+                    } else if (/^og:(?:url|image|audio|video)(?:(?:$|:)(?:url|secure_url|$))/.test(node.getAttribute('property')) && node.hasAttribute('content')) {
+                        if (this._isRelationUrl(node.getAttribute('content'))) {
+                            addOutgoingRelation(new AssetGraph.HtmlOpenGraph({
+                                from: this,
+                                to: {
+                                    url: node.getAttribute('content')
+                                },
+                                node: node,
+                                ogProperty: node.getAttribute('property')
+                            }));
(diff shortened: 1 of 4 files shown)


papandreou:
> Should we (can we) get the content type from a sibling node of `<meta property="og:image:type" content="image/jpeg" />` ?

Oh dear. I guess that would be the right thing to do -- but I guess we can get away with relying on the existing extension/Content-Type based detection as well.

> Should each type and sub type have their own relation instead of them all sharing the same and relying on ogProperty to figure out what property to set?

I don't see much of a drawback to keeping them as one type rather than four, so I'm cool with doing that. We can revisit if it turns out to be a problem.

Munter:
I don't have any big ambitions around automatically generating these relations with a transform. For now I'm satisfied with just picking them up correctly, which the current state covers. I'll merge

papandreou:
:+1:

PR #623, opened 2016-09-16
Add relation/HtmlServiceWorkerRegistration. Closes #493

### LINKED ISSUE #493: Service workers: Model <link rel=serviceworker href=...>
https://twitter.com/addyosmani/status/722798525197770752

@Munter, I couldn't tell whether this is already covered by your WIP.

Munter:


papandreou:
Needs to be integrated with `bundleSystemJs` as well to support detection of entry points inside the service worker. Should just be a matter of adding 'HtmlServiceWorkerRegistration' alongside 'JavaScriptWebWorker':

https://github.com/assetgraph/assetgraph/blob/master/lib/transforms/bundleSystemJs.js#L13
https://github.com/assetgraph/assetgraph/blob/master/lib/transforms/bundleSystemJs.js#L646
https://github.com/assetgraph/assetgraph/blob/master/lib/transforms/bundleSystemJs.js#L730
https://github.com/assetgraph/assetgraph/blob/master/lib/transforms/bundleSystemJs.js#L868

... but it'd be nice with a test as well :)

Munter:
Can SystemJS be used inside workers? Isn't this where you would just use `importScripts()` ?

papandreou:
Yes, it can. `importScript` is the equivalent of `<script src=...>` so it leaves a lot to be desired feature-wise. A while ago I added support for picking up System.js usage in Web Workers so the `System.import(...)` calls found in workers are treated as "first class" endpoints that are traced along with other endpoints found in `<script>`, and the resulting bundles are loaded via `importScript`.

papandreou:
https://github.com/assetgraph/assetgraph/blob/master/test/transforms/bundleSystemJs.js#L314-L325

papandreou:
Fixed in 86279bf0

Munter:
Thanks

PR #626, opened 2016-09-20
Add ability to handle canonical relative urls

Munter:
This adds a new configuration for the assetgraph constructor: `canonicalRoot`
A new `relation.hrefType: 'canonical'`has also been added.

The idea is to let assetgraph populate any href to a URL that is inside the `canonicalRoot` as if it was inside `assetGraph.root`.

This will enable us to pick up absolute URL's, like you often see in blog setups like jekyll or wordpress, where the deployment domain is prepended to the url. The same problem goes for OpenGraph relations, where the specification says to use an absolute url, but the assets are locally available.

Am I even using canonical correctly at this point? Should this be called `deploymentRoot` and `deploymentRelative` instead?

Work in progress...

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -59,6 +59,7 @@ function AssetGraph(options) {
 
     // this.root might be undefined, in which case urlTools.urlOrFsPathToUrl will use process.cwd()
     this.root = normalizeUrl(urlTools.urlOrFsPathToUrl(this.root, true)); // ensureTrailingSlash
+    this.canonicalRoot = normalizeUrl(urlTools.urlOrFsPathToUrl(this.canonicalRoot, true)); // ensureTrailingSlash
 
     this._assets = [];
     this._relations = [];
--- a/lib/index.js
+++ b/lib/index.js
@@ -51,15 +51,34 @@ var util = require('util'),
  * @api public
  */
 function AssetGraph(options) {
+    options = options || {};
+
     if (!(this instanceof AssetGraph)) {
         return new AssetGraph(options);
     }
     EventEmitter.call(this);
+
+    if ('canonicalRoot' in options) {
+        if (typeof options.canonicalRoot !== 'string') {
+            throw new Error('AssetGraph: options.canonicalRoot must be a URL string');
+        }
+
+        // Validate that the given canonicalRoot is actually a URL
+        // regexes lifted from one-validation fragments.domainPart and fragments.tld
+        if (!(/^(?:https?:)?\/\/[a-z0-9](?:[\-a-z0-9]*[a-z0-9])?\.[a-z][\-a-z]*[a-z]/i).test(options.canonicalRoot)) {
+            throw new Error('AssetGraph: options.canonicalRoot must be a URL string');
+        }
+
+        // Ensure trailing slash on canonical root
+        if (options.canonicalRoot[options.canonicalRoot.length - 1] !== '/') {
+            options.canonicalRoot += '/';
+        }
+    }
+
     _.extend(this, options);
 
     // this.root might be undefined, in which case urlTools.urlOrFsPathToUrl will use process.cwd()
     this.root = normalizeUrl(urlTools.urlOrFsPathToUrl(this.root, true)); // ensureTrailingSlash
-    this.canonicalRoot = normalizeUrl(urlTools.urlOrFsPathToUrl(this.canonicalRoot, true)); // ensureTrailingSlash
 
     this._assets = [];
     this._relations = [];
--- a/lib/index.js
+++ b/lib/index.js
@@ -383,6 +383,13 @@ _.extend(AssetGraph.prototype, {
                 url = assetConfig.url = that.resolveUrl(fromUrl, url);
             }
 
+            if (that.canonicalRoot) {
+                if (url.indexOf(that.canonicalRoot) === 0) {
+                    url = url.replace(that.canonicalRoot, that.root);
+                    assetConfig.url = url;
+                }
+            }
+
             var protocol = url.substr(0, url.indexOf(':')).toLowerCase();
             if (protocol === 'javascript') {
                 assetConfig.text = url.replace(/^javascript:/i, '');
--- a/lib/relations/Relation.js
+++ b/lib/relations/Relation.js
@@ -111,6 +111,8 @@ Relation.prototype = {
                 href = urlTools.buildRootRelativeUrl(this.baseUrl, targetUrl, this.assetGraph && this.assetGraph.root);
             } else if (hrefType === 'relative') {
                 href = urlTools.buildRelativeUrl(this.baseUrl, targetUrl);
+            } else if (hrefType === 'canonical' && this.assetGraph) {
+                href = targetUrl.replace(this.assetGraph.root, this.assetGraph.canonicalRoot);
             } else if (hrefType === 'protocolRelative') {
                 href = urlTools.buildProtocolRelativeUrl(this.baseUrl, targetUrl);
             } else {
@@ -177,6 +179,8 @@ Relation.prototype = {
                 this._hrefType = 'protocolRelative';
             } else if (/^\//.test(href)) {
                 this._hrefType = 'rootRelative';
+            } else if (this.assetGraph && this.assetGraph.canonicalRoot && href.indexOf(this.assetGraph.canonicalRoot) === 0) {
+                this._hrefType = 'canonical';
             } else if (/^[a-z\+]+:/i.test(href)) {
                 this._hrefType = 'absolute';
             } else {
--- a/lib/relations/Relation.js
+++ b/lib/relations/Relation.js
@@ -111,8 +111,6 @@ Relation.prototype = {
                 href = urlTools.buildRootRelativeUrl(this.baseUrl, targetUrl, this.assetGraph && this.assetGraph.root);
             } else if (hrefType === 'relative') {
                 href = urlTools.buildRelativeUrl(this.baseUrl, targetUrl);
-            } else if (hrefType === 'canonical' && this.assetGraph) {
-                href = targetUrl.replace(this.assetGraph.root, this.assetGraph.canonicalRoot);
             } else if (hrefType === 'protocolRelative') {
                 href = urlTools.buildProtocolRelativeUrl(this.baseUrl, targetUrl);
             } else {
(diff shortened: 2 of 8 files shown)


papandreou:
Adding `canonicalRoot` to the AssetGraph instance is a good idea, but I'm not sure introducing a new `hrefType` will cover all the cases because it also implies "absolute"? It seems more like a separate thing. Maybe it can be built right into the `crossorigin` getter? I'm thinking it could evaluate to true if the relation points outside `assetGraph.root` AND outside `assetGraph.canonicalRoot`, if present?

lib/index.js
```diff
@@ -59,6 +59,7 @@ function AssetGraph(options) {
 
     // this.root might be undefined, in which case urlTools.urlOrFsPathToUrl will use process.cwd()
     this.root = normalizeUrl(urlTools.urlOrFsPathToUrl(this.root, true)); // ensureTrailingSlash
+    this.canonicalRoot = normalizeUrl(urlTools.urlOrFsPathToUrl(this.canonicalRoot, true)); // ensureTrailingSlash
```
papandreou: We should probably require `canonicalRoot` to be a url, so `urlTools.urlOrFsPathToUrl` won't help.

Hmm, we could also allow specifying the canonical root as a source expression (https://www.w3.org/TR/CSP3/#grammardef-source-expression, our impl. here: https://github.com/assetgraph/assetgraph/blob/master/lib/matchSourceExpression.js) so you can also express the "both http and https" case. That might help us get rid of some of the special casing of protocol-relative urls when we're populating from `file://`.

Munter:
It's not enough to have it in the `crossorigin` getter though. We should be able to switch relation href rendering by changing the `hrefType` around. This would also simplify the canonical implementation in assetgraph-builder

papandreou:
> We should be able to switch relation href rendering by changing the `hrefType` around

Again, I'm worried about overloading `hrefType` for this purpose -- the switch could also be made by switching around another property, right?

Munter:
So maybe we need a new property, or getter/setter pair. I'd like the signature to end up with this though:

```
Relation {
  href: 'http://canonicalurl.com/foo.js',
  to: {
    url: path.join(assetGraph.root, 'foo.js')
  }
}
```

so, assets resolved from any asset config url or relation href, which matches the canonicalRoot, get resolved as file url's relative to assetgraph.root, but the relation href is kept intact and canonical

papandreou:
The current switch is:

```js
            .if(options.canonicalUrl)
                // Maybe this should be the effect of updating assetGraph.root?
                .moveAssets({isLoaded: true, isInline: false}, function (asset, assetGraph) {
                    if (asset.url.indexOf(assetGraph.root) === 0) {
                        return assetGraph.resolveUrl(options.canonicalUrl, urlTools.buildRelativeUrl(assetGraph.root, asset.url));
                    }
                })
                .updateRelations({from: {type: 'Html', isFragment: true, nonInlineAncestor: {type: ['Rss', 'Atom']}}}, {hrefType: 'absolute'})
            .endif()
```

It's not really bad, is it? Do we really need to add another getter/setter pair just so we can simplify it?

Munter:
I don't understand what the suggestion is then. If not an additional `hrefType` then what?

papandreou:
My suggestion is to expand the `crossorigin` getter to also be aware of `assetGraph.canonicalRoot` and to return false for relations from within `assetGraph.root` to within `assetGraph.canonicalRoot` (and possibly vice versa).

But then you said:

> It's not enough to have it in the crossorigin getter though. We should be able to switch relation href rendering by changing the hrefType around. This would also simplify the canonical implementation in assetgraph-builder

... and maybe I didn't quite understand what you meant by "switch relation href rendering", but I assumed it referred to that snippet of buildProduction code that I pasted? If it's only about rewriting the `url` properties of assets within `assetGraph.root` at some point in `buildProduction`, I think the current snippet is OK.

Munter:
It's about being able to discover local assets in the population phase, based on an absolute or protocol relative url matching the canonicalRoot. This has to be a part of the assetgraph internals and not a transform after thought as part of a build system

papandreou:
Ah, that mission statement makes a lot of sense. And since you'll never have explicit `file://`... urls in your app in development, I guess there really doesn't need to be an explicit rewrite step if we're able to pull this off.

I think it might be achievable without introducing that `hrefType` of canonical. How about doing something like this at the start of `refreshHref`:

```js
var targetUrl = this.to && this.to.url;
var baseUrl = this.baseUrl;
var canonicalRoot = this.assetGraph && this.assetGraph.canonicalRoot;
var root = this.assetGraph && this.assetGraph.root;
if (canonicalRoot && root) {
    if (baseUrl.indexOf(root) === 0) {
        baseUrl = baseUrl.replace(root, canonicalRoot);
    }
    if (targetUrl.indexOf(root) === 0) {
        targetUrl = targetUrl.replace(root, canonicalRoot);
    }
}
```

and then in the rest of the function use `baseUrl` instead of `this.baseUrl` and `canonicalRoot || root` instead of `this.assetGraph && this.assetGraph.root`?

lib/index.js
Munter: Is there a limited source-expression tester available somewhere? I do want the host part, but don't really care for the pure scheme, nonce or digest validation

papandreou: Hmm, not that I'm aware of. Come to think of it, would we want to support anything but the below?

* `http://example.com`
* `https://example.com`
* `example.com`
* `example.com/foo/bar/`

Munter: Those are the only examples I could come up with as well. Possibly with the `//` prefix on the domain.

I'm still a bit unsure if supporting all of these syntaxes is preferable over just a simple substring match. If we add the source expression match I think it's important that we can make fully reversible operations if a relation switches from canonical to local, back to canonical. In that situation I would prefer the href to end up exactly as originally written

papandreou: I agree, let's start out by requiring the canonical url to be absolute.

Munter: I replaced this with a URL validation of the input options

Munter:
@papandreou I tried to remove the `canonical` hrefType. Now replaced it with a `canonical` getter/setter pair, so it's possible to keep the intention of whether a relation should be canonical or not, even when files move around. There's one thing I'm not so happy about, I bet you'll spot it immediately

lib/relations/Relation.js
```diff
@@ -148,6 +155,36 @@ Relation.prototype = {
...
+        return this._canonical;
+    },
+
+    set canonical(isCanonical) {
+        this._canonical = !!isCanonical;
+
+        var hrefTypeRoot = this._canonical ? this.assetGraph.canonicalRoot : '';
+        // We need to update the hrefType manually here because otherwise hrefType in refreshHref
+        // will be based on the old href value, not the new one
+        this._hrefType = Object.getOwnPropertyDescriptor(Relation.prototype, 'hrefType').get.call({ href: hrefTypeRoot });
```
papandreou: This shouldn't be necessary as `hrefType` only expresses the intention to use eg. a relative href if possible -- `refreshHref` will still issue an absolute href if the host names differ.

If the `refreshHref` below doesn't do the right thing without the above two lines, I would say that it has a bug.

test/constructor.js
```diff
@@ -0,0 +1,67 @@
...
+            }, 'to throw', /canonicalRoot must be a URL string/);
+        });
+
+        it('should throw when canonicalRoot is not a URL', function () {
+            return expect(function () {
+                return new AssetGraph({ canonicalRoot: 'foo' });
+            }, 'to throw', /canonicalRoot must be a URL string/);
+        });
+
+        it('should accept a http URL', function () {
```
papandreou: an

lib/index.js
```diff
@@ -51,10 +51,30 @@ var util = require('util'),
...
     EventEmitter.call(this);
+
+    if ('canonicalRoot' in options) {
+        if (typeof options.canonicalRoot !== 'string') {
+            throw new Error('AssetGraph: options.canonicalRoot must be a URL string');
+        }
+
+        // Validate that the given canonicalRoot is actually a URL
+        // regexes lifted from one-validation fragments.domainPart and fragments.tld
+        if (!(/^(?:https?:)?\/\/[a-z0-9](?:[\-a-z0-9]*[a-z0-9])?\.[a-z][\-a-z]*[a-z]/i).test(options.canonicalRoot)) {
```
papandreou: No port allowed?

test/relations/Relation.js
```diff
@@ -61,6 +66,148 @@ describe('relations/Relation', function () {
...
+                    crossorigin: false,
+                    href: 'http://canonical.com/local.js',
+                    to: {
+                        url: 'file://' + pathModule.join(testDataDir, 'local.js')
+                    }
+                });
+            });
+        });
+
+        it('should remove the canonical root fromthe href of a local file', function () {
```
papandreou: from the

lib/relations/Relation.js
Munter: Then we have a bug. The last test case I added shows it if this line is commented out

papandreou:
This looks really promising! I'm not sure when I'd actually use the `canonical` setter -- is that mostly there for completeness?

Munter:
It's to give you the ability to toggle canonicalness (I think I just made that word up) instead of having to manually set the href by concatenating your canonical root and the assetgraph root relative url to the asset. We already have this use case in assetgraph-builder where we update some links. Also I can imagine some transforms that might switch canonical url's to root relative, for example when you want to build the output of certain CMS'es that just love to add canonical url's all over the place, leaking configuration complexity into your server setups

lib/relations/Relation.js
papandreou: Right, I see it now. Because "http://canonical.com/local.js" is an absolute url, the relation will get an `_hrefType` of `absolute` as soon as something accesses the `hrefType` getter. That remains unaffected by flipping `_canonical` to false, so `refreshHref` will issue an absolute `file://` url in that test.

What you want is to automatically switch to a "lesser" type of href if possible when the canonicalness is switched off so you won't end up with eg. an absolute `file://` url. I guess that makes sense -- but we don't really know if a `hrefType` of `relative` would be sufficient as the user hasn't really expressed the intent through that originally absolute href. Applications that use pushState generally need at least `rootRelative` hrefs.

papandreou: https://github.com/assetgraph/assetgraph/pull/634

Munter: Awesome!

papandreou:
I'm really curious about what the companion changes to ag-b are going to look like, but these changes look good to me in this state.

:+1:

Munter:
I'll fix the nits and merge when I get home

lib/index.js
Munter: Port is allowed. I don't look at anything beyond the tld. Added a test for it

Munter:
Nits fixed. Rebased and pushed. Merging on green

PR #640, opened 2016-10-20
Don't include a document fragment in the asset url, treat it as part of the relation href

papandreou:
As discussed on gitter tonight.

// cc @Munter

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -401,6 +401,7 @@ _.extend(AssetGraph.prototype, {
             if (!/^[a-zA-Z-\+]+:/.test(url)) {
                 url = assetConfig.url = that.resolveUrl(fromUrl, url);
             }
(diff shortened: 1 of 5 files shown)


Munter:
Love it! Will try it out linked into hyperlink tonight before I merge. Should cut down the amount of work done considerably

Munter:
Published in 3.0.0-35

papandreou:
Hmm, that reminds me that we skip relations that only consist of a `#fragment`. Seems like that new hyperlink feature is a good occassion to stop doing that so fragment links to thepage itself can also be checked.

papandreou:
> ... a good occasion to stop doing that so fragment links to the page itself can also be checked

https://github.com/assetgraph/assetgraph/pull/642

PR #642, opened 2016-10-22
Model (self-)relations whose href is the empty string or #fragment

papandreou:


Munter:
This is great. Less special casing, which makes the api more intuitive!

papandreou:
I agree. `_isRelationUrl` used to exclude other things such as`javascript:`, but we model those as well now :)

papandreou:
Okay, pulling it in.

Munter:
I've been testing it against updated hyperlink code during the day. Everything is better with this patch :D

papandreou:
Great to hear that :)

PR #652, opened 2016-10-30
Make relations owned by the asset containing them

papandreou:
This needs some more polish, but shows the general idea. ag-b tests don't pass yet.

Preserves the `findRelations` API (for now).

CC: @Munter

### DIFF
--- a/lib/assets/Asset.js
+++ b/lib/assets/Asset.js
@@ -575,10 +575,12 @@ extendWithGettersAndSetters(Asset.prototype, {
     },
 
     removeRelation: function (relation) {
-        var outgoingRelations = this.outgoingRelations;
-        var i = outgoingRelations.indexOf(relation);
-        if (i !== -1) {
-            outgoingRelations.splice(i, 1);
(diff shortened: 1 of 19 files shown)


Munter:
Yay! This looks really good

papandreou:
Thanks! I'll try to finish it up ASAP.

It unlocks a bunch of interesting things that we might want to explore right away:

1) New query syntax -- something jquery-ish with the transform pipeline built-in?
2) Something like `markDirty` that also invalidates the relations so they're re-discovered on demand? (We talked about this on the gitter channel about a month ago)
3) Instantiate `Asset` objects right away instead of keeping "asset configs" around, then subclass when the metadata (especially the type) is known. Could clear up a lot of inconsistencies.

Munter:
Those are big wins I think. Will make the behaviour less surprising

papandreou:
Seems like this works now.

Sister PRs:
https://github.com/assetgraph/assetgraph-builder/pull/517
https://github.com/assetgraph/assetgraph-sprite/pull/140

PR #667, opened 2016-12-28
Implement transform to inline critical CSS rules

Munter:
This transform looks for all `Html`-assets that have their first `<link rel="stylesheet">` inside the `<body>`-element.

For each of these pages it collects all DOM-elements that appear on the page before the first `<link rel="stylesheet">`. These are the DOM-elements that are treated as critical to have styled immediately when the brower does its initial render flush.

For each of the critical DOM-nodes it finds the CSS selectors in the pages dependency graph that apply to the node.

CSS selectors matching critical DOM-nodes are then cloned into an inline stylesheet in the pages `<head>`, thus ensuring styling on the critical nodes on the browsers first render flush.

This is particularly useful in Edge and Firefox, but will also yield improved results in Chrome according to [Chromes intent to implement parser blocking for external stylesheets in body](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/ZAPP8aTnyn0) and [Jake Archibalds writeup on link in body behavior](https://jakearchibald.com/2016/link-in-body/)

For a developer to trigger critical CSS inlining, all that is needed is to move the first `<link rel="stylesheet">` into the `<body>` at a relevant position, where it will declaratively denote "the fold" with critical content above and less critical content below.

Example output of [mntr.dk](https://mntr.dk) with stylesheets in body and transform applied: http://doubtful-scissors.surge.sh/ (view source)

### DIFF
--- a/lib/relations/HtmlRelation.js
+++ b/lib/relations/HtmlRelation.js
@@ -52,6 +52,13 @@ extendWithGettersAndSetters(HtmlRelation.prototype, {
         // Create head node if it's missing
         if (!head) {
             var htmlNode = document.querySelector('html');
+
+            if (!htmlNode) {
+                var err = new Error('HtmlRelation.attachToHead: Could not attach asset ' + this.to.toString() + ' to <head>. Missing <html> and <head>');
+                err.asset = asset;
+                throw err;
+            }
+
             head = document.createElement('head');
 
             htmlNode.insertBefore(head, htmlNode.firstChild);
--- a/lib/relations/HtmlStyle.js
+++ b/lib/relations/HtmlStyle.js
@@ -61,6 +61,20 @@ extendWithGettersAndSetters(HtmlStyle.prototype, {
         return this;
     },
 
+    attachToHead: function (asset, position, adjacentNode) {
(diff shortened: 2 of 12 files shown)


papandreou:
> are then cloned

Cloned, not moved? Is that to avoid breaking the cascade?

Munter:
> Cloned, not moved? Is that to avoid breaking the cascade?

Partially. I also couldn't predict all the different cases that might occur when identical CSS across multiple pages suddenly becomes non-identical because different pages have different DOM above the fold. It would potentially void caches across page loads.

It's simply to avoid some complexity for now.

It might be interesting to implement the move behavior. I guess that would also mean the transform would have to ensure the external CSS asset has no other incoming relations, and if it does, clone the entire asset to make it unique for the page. Then run a deduplication of all the newly created CSS assets at the end of the transform. Would that cover the use case aproximately?

papandreou:
> Would that cover the use case approximately?

I think it would, but I don't consider it a prereq for merging this. It might not be worth the added complexity at all.

Was mostly asking out of curiosity :)

PR #710, opened 2017-03-22
Fix relation canonical getter

Munter:
`hrefObj.path.indexOf(canonicalRootObj.path)` would throw when a `mailto:`-link had a host name identical to the `canonicalRoot` because `hrefObj.path === null`.

The fix checks that `hrefObj` has `slashes === true` and that the protocol must be either `http`, `https` or `null`(protocol-relative)

### DIFF
--- a/lib/relations/Relation.js
+++ b/lib/relations/Relation.js
@@ -168,7 +168,7 @@ Relation.prototype = {
                 var hrefObj = urlModule.parse(this.href, false, true);
 
                 canonical = hrefObj.slashes === true
-                    && ['http', 'https', null].indexOf(hrefObj.protocol)
+                    && ['http:', 'https:', null].indexOf(hrefObj.protocol) !== -1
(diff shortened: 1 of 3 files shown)


lib/relations/Relation.js
```diff
@@ -167,7 +167,9 @@ Relation.prototype = {
                 var canonicalRootObj = urlModule.parse(this.assetGraph.canonicalRoot, false, true);
                 var hrefObj = urlModule.parse(this.href, false, true);
 
-                canonical = canonicalRootObj.host === hrefObj.host
+                canonical = hrefObj.slashes === true
+                    && ['http', 'https', null].indexOf(hrefObj.protocol)
```
papandreou: Should be `['http', 'https', null].indexOf(hrefObj.protocol) !== -1` right?

Munter: Right. The test passes because the `slashes` verification before

PR #712, opened 2017-03-24
ES6 support

papandreou:
Seems like it's quite easy to allow ES6 syntax in JavaScript assets by switching esprima into "module" mode. I hope there are no downsides :)

It's also able to parse JSX, although escodegen doesn't support serializing it. Still, better than nothing.

### DIFF
--- a/lib/assets/Html.js
+++ b/lib/assets/Html.js
@@ -799,7 +799,10 @@ extendWithGettersAndSetters(Html.prototype, {
                         var javaScriptObjectLiteral = '({' + node.getAttribute(attributeName).replace(/^\s*\{(.*)\}\s*$/, '$1') + '});',
                             parseTree = null; // Must be set to something falsy each time we make it here
                         try {
-                            parseTree = esprima.parse(javaScriptObjectLiteral);
+                            parseTree = esprima.parse(javaScriptObjectLiteral, {
+                                sourceType: 'module',
+                                jsx: true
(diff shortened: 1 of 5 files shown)


Munter:
Wow, that was easy :)

I guess a whole new bunch of possibilities open up now with import statements and jsx html-tag attributes as relations

papandreou:
> I guess a whole new bunch of possibilities open up now with import statements

Yes, it does. I wonder how we could approach that.

> and jsx html-tag attributes as relations

Hmm, yeah, theoretically, I'm not sure there's a workflow in existence where assetgraph would end up seeing the untranspiled JSX, except by following source map relations. What do you have in mind?

papandreou:
Wrt. reserializing JSX: https://github.com/estools/escodegen/issues/333

Munter:
> Hmm, yeah, theoretically, I'm not sure there's a workflow in existence where assetgraph would end up seeing the untranspiled JSX, except by following source map relations. What do you have in mind?

I think you are right. I'm firmly of the belief that transpiling should happen before any bundler even touches the code, so I would actually hope never to see it. I'd rather pour lots of energy into implement picking up html relations from the resulting ReactDOM constructor calls, if we know we can reason enough about them to statically analyse

papandreou:
I'll merge this, then we can pick up the JSX reserialization thing later.

PR #715, opened 2017-03-27
When esprima.parse(text, {..., sourceType: 'module'}) fails, try falling back to sourceType: 'script'

papandreou:


papandreou:
It sounds like webpack does exactly the same :)

And since we're not evaluating the code, only parsing it, we're not really vulnerable to all the gotchas.

What do you think, @Munter? It fixes a case I ran into where `await` was being used as a variable name in some legacy code. It probably also fixes parsing of anything that uses `with`.

PR #716, opened 2017-03-27
bundleWebpack: Only build the bundles that are referenced by relations

papandreou:
The idea is to avoid doing a bunch of work that's not needed in projects where everything isn't being built in a single `buildProduction` run.

// CC: @gustavnikolaj


Munter:
I think you need to test if this would work with the commonchunksplugin or split points with `System.import(...).then(...)`

papandreou:
I think that would work fine, as you wouldn't explicitly list those common/lazy bundles in `entry` in your config? @gustavnikolaj?

papandreou:
I guess the title should have been "only build the *entry* *points*...".

gustavnikolaj:
Common chunks would not really be common as we would build each entry point individually, but it would work just fine as those are derived from the common chunks between the entrypoints and not listed as entry points themselves, as far as I know.

There will be dangling relations to the common chunks file and that would be satisfied after building the file. But a question remains; would the common file be generated when there is only a single entrypoint - which would probably be the case for us all the time?

https://webpack.js.org/plugins/commons-chunk-plugin/

The commons chunk plugin is most definitely a problem with our approach of building a single entry point at a time...

papandreou:
@gustavnikolaj our === One.com's here?

gustavnikolaj:
@papandreou yes... I guess it would work fine if we didn't call buildProduction independently for each html file.

papandreou:
So cool to merge?

gustavnikolaj:
I guess we should add a test with `buildProduction` running two html files, with a common chunk referenced from each and the plugin enabled in the webpack config.

We can try it out tomorrow at the office?

papandreou:
Sure, man :)

Munter:
I was hoping this transform would be able to be used by al webpack users. But I can see there are already a bunch of assumptions about webpack configuration in the code, which would most likely make things break for "normal" webpack configurations with CommonChunksPlugin and HtmlWebpackPlugin

papandreou:
Sorry, what do you mean? What are those assumptions?

Munter:
Things like injecting CSS relations in the html, that's configured manually by most webpack users, either by not dynamically naming the resulting bundle, or using HtmlWebpackPlugin to dynamically inject a link and script tags at build time. It means that users are used to manually assembling things into a consistent dependency graph themselves (though often failing), and will assume tools to integrate with that.

I do think that most of what these plugins do are better handled by assetgraph, but alas, that message works badly in marketing

papandreou:
Okay, but that's unrelated to this PR, right?

We took a look at HtmlWebpackPlugin, and as it just emits the HTML file as yet another asset, it seems that it wouldn't be impossible to support it. It might require us to change the `buildProduction` CLI parameters/semantics so that you don't even need an "initial" HTML asset to get started.

@gustavnikolaj suggests making a separate `buildWebpack` command for that to keep things focused.

In any case, we should discuss this in a different issue.

Munter:
I guess it doesn't affect this PR in particular, now that I look closer. But we should see if we can get some buy-in from Sean or Jan Timon so they can feed us back information on how to best integrate with what most people do with webpack currently

papandreou:
Getting webpack people involved would be a good idea. At least to the point where they're roughly aware of what we're doing :)

PR #717, opened 2017-03-28
Support webpack code splitting

papandreou:
- [x] Built on top of #716, await that before merging :)

### DIFF
--- a/lib/transforms/bundleWebpack.js
+++ b/lib/transforms/bundleWebpack.js
@@ -168,9 +168,114 @@ module.exports = function (queryObj, options) {
                         url: assetInfos.map(function (url) { return url.url; })
                     }).forEach(function (asset) {
                         var replacementsMade = false;
+                        var requireEnsureAstTemplate;
                         estraverse.traverse(asset.parseTree, {
                             enter: function (node) {
-                                if (node.type === 'FunctionExpression' &&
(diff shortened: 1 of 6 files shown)


papandreou:
This is quite fragile as it depends on very specific code in the emitted bundles. However, users will expect code splitting to work, so IMO the best we can do is to add something like this, and then be careful to test with new webpack versions (thanks, Greenkeeper).

... And hope that https://github.com/h5bp/lazyweb-requests/issues/201 leads somewhere.

Munter:
I just looked through webpacks examples for code splitting using `import()`, and it actually looks as if the compiled output is using the same `requireEnsure` function that this PR covers!

https://github.com/webpack/webpack/search?p=1&q=requireEnsure&type=&utf8=%E2%9C%93

Munter:
Confirmed by sokra. All async loading syntaxes compile down to the same thing :D

Munter:
We might want to consider moving these tree traversals into the `JavaScript` asset if we want to support a use case where the builder is run on webpack output from disk

papandreou:
> it actually looks as if the compiled output is using the same `requireEnsure` function that this PR covers!

That's what we hoped, but didn't actually check. That'll sure make the upgrade to webpack 2 easier, if there's even anything to do.

> We might want to consider moving these tree traversals into the `JavaScript` asset

Good idea. It would also be quite easy to put it into a separate `pickUpWebpackRelations` transform.

PR #723, opened 2017-03-28
Webpack 2 compatibility

papandreou:


papandreou:
Something's up with the file-loader.

Repro: `mocha --no-exit -g 'bundleWebpack.*discover relations|should work when combined with the asset list plugin'`

papandreou:
`mocha -i -g bundleSystemJs` passes. Must be some interference from systemjs-builder.

papandreou:
@guybedford, do you think there would be a way for systemjs-builder to clean up?

lib/transforms/bundleSystemJs.js
```diff
@@ -928,9 +939,14 @@ module.exports = function (options) {
             bundleStrategyOptions.conditions = options.conditions;
         }
 
+        var globalSnapshotBeforeBundling = Object.assign({}, global);
+        _.defaults(global, systemJsGlobals);
         return bundleStrategy(builder, entryPointNames.map(function (entryPointName) {
             return { name: entryPointName };
         }), bundleStrategyOptions).then(function (bundles) {
+            Object.keys(systemJsGlobals).forEach(function (key) {
+                delete global[systemJsGlobals];
```
gustavnikolaj: s/systemJsGlobals/key/

papandreou: Whoops, that shouldn't even be there, the cleanup is handled in the `finally` block. Fixed.

guybedford:
This seems sensible! Will definitely aim to remove leaks in subsequent versions.

papandreou:
Landed in 3.0.0-48 / ag-b 5.0.0-38

PR #728, opened 2017-04-10
Insert HtmlScript bundles at the point of the last asset in the bundl…

Munter:
…e rather than the first asset. Ensures scripts are not moved before DOM they depend on

Example that would fail before:

```html
<script>console.log('foo')</script>
<h1>Hello World</h1>
<script>console.log(document.querySelector('h1').innerText</script>
```


Munter:
I don't understand why Travis is failing. All tests run without error on my machine

PR #741, opened 2017-05-12
Add transform that uses sw-precache to generate a service worker

papandreou:
- [x] Awaiting https://github.com/GoogleChrome/sw-precache/pull/297

Sister PR: https://github.com/assetgraph/assetgraph-builder/pull/531


Munter:
This seems to always create a serviceworker per html-page. We should consider an option to have a single serviceworker for the whole graph

lib/transforms/addPrecacheServiceWorker.js
```diff
@@ -0,0 +1,69 @@
...
+                throw new Error('addCacheManifestSinglePage: Cannot generate a CacheManifest for an inline Html asset');
+            }
+            var dynamicUrlToDependencies = {};
+            var assetByUrl = {};
+            assetGraph.eachAssetPostOrder(htmlAsset, {type: query.not(['HtmlAnchor', 'SvgAnchor', 'HtmlMetaRefresh', 'HtmlCacheManifest', 'HtmlConditionalComment', 'JavaScriptSourceUrl', 'JavaScriptSourceMappingUrl', 'JavaScriptSourceMap'])}, function (asset) {
+                // But only if the asset isn't inline, has been loaded, and isn't already in the manifest:
+                if (!asset.isInline && asset.isLoaded && asset !== htmlAsset) {
+                    var url = require('urltools').buildRootRelativeUrl(assetGraph.root, asset.url, assetGraph.root);
+                    assetByUrl[url] = asset;
+                    dynamicUrlToDependencies[url] = asset.isText ? asset.text : asset.rawSrc;
```
papandreou: It mirrors the config property accepted by sw-precache, but I guess you're right that it sounds a bit weird in this context :)

lib/transforms/addPrecacheServiceWorker.js
```diff
@@ -0,0 +1,69 @@
...
+                    logger: function (message) {
+                        assetGraph.emit('info', htmlAsset.urlOrDescription + ': ' + message);
+                    }
+                }, cb);
+            }).then(function (serviceWorkerStr) {
+                serviceWorkerStr = serviceWorkerStr.replace(/^(var precacheConfig = )(\[[^\n]*\]);/m, function ($0, before, urlAndHashPairs) {
+                    return before + '[' + JSON.parse(urlAndHashPairs).map(function (urlAndHash) {
+                        return '[' + JSON.stringify(urlAndHash[0]) + '.toString(\'url\'),' + JSON.stringify(urlAndHash[1]) + ']';
+                    }).join(',') + '];';
+                });
```
papandreou: Yes, but we need to rewrite the url strings to `'...'.toString('url')` for AssetGraph to be able to pick up the references and update them if they're moved later.

lib/transforms/addPrecacheServiceWorker.js
```diff
@@ -0,0 +1,69 @@
...
+                var serviceWorkerUrl = htmlAsset.url.replace(/(?:\.\w+)?([?#]|$)/, '-service-worker.js');
+                var serviceWorker = new assetGraph.JavaScript({
+                    url: serviceWorkerUrl,
+                    text: serviceWorkerStr
+                });
+                serviceWorker.outgoingRelations.forEach(function (relation) {
+                    relation.to = assetByUrl[relation.href];
+                });
```
papandreou: Good point. I don't even remember why (I thought) it was necessary. I'll take another look.

lib/transforms/addPrecacheServiceWorker.js
```diff
@@ -0,0 +1,69 @@
...
+                    text:
+                        'if (\'serviceWorker\' in navigator) {\n' +
+                        '    window.addEventListener(\'load\', function () {\n' +
+                        '        navigator.serviceWorker.register(\'' + urlTools.buildRootRelativeUrl(assetGraph.root, serviceWorkerUrl, assetGraph.root).replace(/'/g, '\\\'') + '\');\n' +
+                        '    });\n' +
+                        '}\n'
+                });
+                assetGraph.addAsset(serviceWorkerRegistrationScript);
+                serviceWorkerRegistrationScript.populate();
+                serviceWorkerRegistrationScript.outgoingRelations[0].to = serviceWorker;
```
papandreou: Ditto.

lib/transforms/addPrecacheServiceWorker.js
Munter: Ah. Yet another thing that will be fixed with extended modeling of `Serviceworker`-type relations

papandreou: I don't think so as this is basically a var at the top of the generated script:

```js
var precacheConfig = [["/static/foo.js", "d3b07384d113edec49eaa6238ad5ff00"], ["/static/bar.js", "c157a79031e1c40f85931829bc5fc552"]];
```

papandreou:
That's also a great point -- shouldn't be too hard.

lib/transforms/addPrecacheServiceWorker.js
Munter: ah, another abstraction layer. That's too bad

papandreou: Yeah, https://github.com/h5bp/lazyweb-requests/issues/201 again again.

lib/transforms/addPrecacheServiceWorker.js
papandreou: Ah, turns out it's just to avoid running the `populate` transform because we already know exactly which assets will be referenced.

It's still ugly, though. I'll think about it.

papandreou: Adding a comment for now.

lib/transforms/addPrecacheServiceWorker.js
papandreou: Renamed the variable to `textOrRawSrcByUrl` now.

papandreou:
Added a `single:true` mode now where a shared service worker is generated.

Need to find out how to expose that option in buildProduction, otherwise I think this is good to go now.

papandreou:
Come to think of it -- maybe the single precache service worker shared between all the pages would be the least surprising default mode -- we could just expose that in buildProduction. What do you think, @Munter?

papandreou:
There, I added some more refinements and tests. Now it also handles the case where multiple origins are in play (generate one service worker per origin).

PR #750, opened 2017-06-24
es6/2017-ify transforms

papandreou:
I did a sweep through all the transforms (except `loadAssets` and `populate` which are involved in another plan). The changes include:

* `for ... of` instead of `forEach`
* `const`/`let` instead of `var`, changed style to one per line
* arrow functions
* a little bit of destructuring when I noticed it
* `async`/`await`
* template strings

In other words, I move that we drop support for node.js < 8 in the next major release.

### DIFF
--- a/lib/transforms/bundleRelations.js
+++ b/lib/transforms/bundleRelations.js
@@ -1,24 +1,23 @@
-var _ = require('lodash'),
-    urlTools = require('urltools'),
-    postcss = require('postcss'),
-    AssetGraph = require('../AssetGraph'),
-    assetGraphConditions = require('../assetGraphConditions');
+const _ = require('lodash');
+const urlTools = require('urltools');
+const postcss = require('postcss');
+const AssetGraph = require('../AssetGraph');
+const assetGraphConditions = require('../assetGraphConditions');
 
-module.exports = function (queryObj, options) {
+module.exports = (queryObj, options) => {
     options = options || {};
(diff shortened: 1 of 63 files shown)


lib/transforms/bundleRelations.js
```diff
@@ -32,14 +31,14 @@ module.exports = function (queryObj, options) {
...
-                Array.prototype.push.apply(discriminatorFragments, _.map(relation.conditionalComments, 'nodeValue'));
+                discriminatorFragments.push(..._.map(relation.conditionalComments, 'nodeValue'));
             }
             if (relation.node && relation.node.hasAttribute && relation.node.hasAttribute('bundle')) {
                 discriminatorFragments.push(relation.node.getAttribute('bundle'));
             }
             if (relation.type === 'HtmlStyle') {
                 discriminatorFragments.push(relation.node.getAttribute('media') || 'all');
-                for (i = 0 ; i < relation.node.attributes.length ; i += 1) {
+                for (let i = 0 ; i < relation.node.attributes.length ; i += 1) {
```
Munter: `Array.from(...).forEach(...)`

papandreou: Fixed + replaced a bunch of `_.extend`s with `Object.assign`

papandreou:
Okay, I'll go ahead and merge. Sorry in advance if this ends up causing a lot of conflicts with a long-running v3 branch.

PR #754, opened 2017-07-12
More ESification

Munawwar:
- replaces vars with const, string concats with template string and built-in array context with arrow function
- changes some if/else logic to exit early.
- removes typeof x !== 'undefined'. And uses x !== undefined.

### DIFF
--- a/lib/assets/Asset.js
+++ b/lib/assets/Asset.js
@@ -4,7 +4,7 @@
  * An asset object represents a single node in an AssetGraph, but can
  * be used and manipulated on its own outside the graph context.
  */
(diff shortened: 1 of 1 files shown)


papandreou:
Well, thanks. Although the feature/simplifyAddAsset branch is still very much work in progress and is already too far reaching :)

Munawwar:
Oh, build failed
"/home/travis/build/assetgraph/assetgraph/lib/assets/Asset.js
  978:188  error  Parsing error: Unexpected token )"

Munawwar:
Fixed. Raised new PR.

PR #760, opened 2017-07-28
Support :after { content: attr(...); }

papandreou:
Also, don't break when there's no content attribute in a pseudo element rule (produced a styledText entry with text: 'undefined' before).

### DIFF
--- a/test/util/fonts/getTextByFontProperties.js
+++ b/test/util/fonts/getTextByFontProperties.js
@@ -487,16 +487,7 @@ describe('lib/util/fonts/getTextByFontProperties', function () {
                 '<h1></h1>'
             ].join('\n');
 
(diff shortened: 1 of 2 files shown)


test/util/fonts/getTextByFontProperties.js
```diff
@@ -481,6 +481,24 @@ describe('lib/util/fonts/getTextByFontProperties', function () {
...
+            return expect(htmlText, 'to exhaustively satisfy computed font properties', [
+                {
+                    text: '',
+                    props: {
+                        'font-family': 'font1',
+                        'font-weight': 700,
+                        'font-style': 'normal'
+                    }
+                }
+            ]);
```
Munter: I would expect the empty array here, since pseudo elements don't really exist without `content`

Issue #219, opened 2014-08-28
Atom feed not picked up


Munter:
``` html
<link href=http://yeoman.io/blog/atom.xml type=application/atom+xml rel=alternate title="Yeoman Blog Atom Feed">
```

This doesn't get picked up as an Atom feed, but as a generic `Xml`-asset. The standard says the `.xml`-extension is valid, so we should support it.

The `Atom`-asset is also missing some of the improvements we made to the `Rss`-asset with finding outgoing relations in html fragments

Issue #208, opened 2014-07-21
New hreftype: 'inline'


Munter:
When traversing the graph by relation hrefTypes it looks like inlined relations that are `data:`-uris, are actually treated as absolute. That feels unintuitive to me.

How big of an operation would it be to add an inline hreftype for data-uris?

I'm guessing we would double up on some of the work already done in Asset.inline, which might not be desirable

papandreou:
Sounds like a very good idea.

The inline/external state predates the `hrefType`, and for some reason I didn't realize that it'd be much nicer to represent the inline-ness as a distinct `hrefType` value when it was introduced.

PR #761, opened 2017-07-29
Rework AssetGraph#addAsset, explore API tweaks

papandreou:
#755 offers some opportunities to tweak the API and the invariants of the graph. This is a spike that (for now) has a scope of:

- [x] Alias `AssetGraph#addAsset` as `AssetGraph#add` and adopt almost all features from the `loadAssets` transform, including file system path and wildcard support.
- [x] Stop using the asset constructors directly (eg. `new AssetGraph.Html()`), settle the exact scope for `AssetGraph#add` so that it's no longer needed
- [ ] ~~Stop supporting assets that aren't in a graph~~ (some tests rely on this, so we do a best effort, revisit later)
- [ ] ~~~Disallow removing assets that have incoming relations -- leave them in the graph (maybe `.unload()` them instead). The benefit is that relations won't have a "dangling" state where they don't point to an instantiated asset, thus simplifying the model a lot. Peter: Consider just detaching all incoming relations when an asset is removed.~~~ Not totally sure about this one -- I've implemented it and opened a separate PR that we can merge after this one: https://github.com/assetgraph/assetgraph/pull/766
- [x] Add the ability to upgrade an existing asset to a more specific type, eg. Xml to Atom.
- [x] Do away with the concept of "unresolved relations", remove `assetGraph.findRelations(..., true)`
- [x] Try to avoid double implementation of data: url parsing (Asset#load and Asset#populate)
- [x] Enforce that multiple assets never share the same url.
- [x] Add an "I don't care, just make up a unique url within the root for me" mode when adding an asset.
- [x] Warn if an asset has ambiguous types, eg. is used as both a PNG and a JavaScript.
- [x] Make an `AssetGraph#addAll` that returns an  #array
- [x] Include the fact that we followed a specific relation type when discovering an asset in the type heuristic
- [x] Consider moving the error reporting from populate into `Asset#load`
- [x] Support `asset.addRelation({ ..., href: ... });` as an alternative to the more verbose `asset.addRelation({ ..., to: { url: ...  }});`
- [x] Consider letting `asset.addRelation` auto-create the node if one isn't specified in the options object, thus taking over the responsibility from the current `Relation#attach`.
- [x] Support media attribute when instantiating `HtmlStyle` and `CssImport`
- [x] Allow assets to define new `position` values for `addRelation`, eg. `lastInHead` or `firstInBody` for Html.
- [x] Add support for `hrefType = 'inline'`, also when adding a fresh relation (#208)
- [x] Rename Asset#load to #loadAsync

Fixes https://github.com/assetgraph/assetgraph/issues/219, https://github.com/assetgraph/assetgraph/issues/208


papandreou:
More things to consider:

* Should we rename `AssetGraph#add` back to `addAsset` now that `addAll` and `add` are separate? It was mostly the singular/plural thing that made me prefer `add` in the first place. Otherwise I assume we'll also have to rename `AssetGraph#removeAsset` to `remove`.
* Should we turn `Relation#to` into a setter so you don't have to call `Relation#refreshHref` after updating it?

PR #762, opened 2017-08-10
Switch AssetGraph#_assets over to a Set instance

papandreou:
Not sure we want to do this, but interesting that the change was that simple. Seems to save a second or two when running the test suite.


Munter:
Are there any other advantages than avoiding messing with array splicing? Seems like it's pretty well hidden under abstraction layers. I like the Set version better though

papandreou:
No, the main win is avoiding Array#splice and thus (maybe) gaining a little bit of performance. It's just an implementation detail, so we can change our minds back if we want :)

PR #763, opened 2017-08-12
Snap to available properties

Munter:


lib/util/fonts/snapToAvailableFontProperties.js
```diff
@@ -0,0 +1,127 @@
...
+};
+
+var INITIAL_VALUES = {
+    'font-family': undefined,
+    'font-stretch': 'normal',
+    'font-weight': 400,
+    'font-style': 'normal'
+};
+
+function ascendng(a, b) {
```
papandreou: ascending

lib/util/fonts/snapToAvailableFontProperties.js
```diff
@@ -0,0 +1,127 @@
...
+    });
+    propsToSnap = _.defaults({}, propsToSnap, INITIAL_VALUES);
+
+    // System font, we can't know about the full properties. Early exit
+    if (typeof propsToSnap['font-family'] === 'undefined') {
+        return propsToSnap;
+    }
+
+    // Match font-family first
+    var familymatches = _.filter(fontFaceDeclarations, { 'font-family': propsToSnap['font-family'] });
```
papandreou: Should probably be camel cased: `familyMatches`

lib/util/fonts/resolveFontWeight.js
```diff
@@ -1,44 +1,33 @@
...
-            if (possibleWeight > 500) {
-                value = possibleWeight;
-            } else if (!value) {
-                value = possibleWeight;
-            }
-        });
-
-        return value;
-    }
+    return searchOrder[0];
```
papandreou: Looks like this code could be optimized quite a bit -- there's a bunch of intermediate arrays that get thrown away. OTOH it's very clear what's going on, so if it's not too hot we should probably leave it like this.

lib/util/fonts/snapToAvailableFontProperties.js
```diff
@@ -0,0 +1,127 @@
...
+    var stretchStartIndex = fontStretchValues.indexOf(propsToSnap['font-stretch']);
+    var stretchGroups = _.groupBy(familymatches, 'font-stretch');
+
+    var firstHalf, lastHalf, stretchSearchOrder;
+    var stretchMatches = [];
+
+    if (stretchStartIndex <= fontStretchValues.indexOf('normal')) {
+        // When value is 'normal' or lower, check denser values first, then less dense
+        firstHalf = fontStretchValues.slice(0, stretchStartIndex + 1);
+        lastHalf  = fontStretchValues.slice((stretchStartIndex + 1) - fontStretchValues.length);
```
papandreou: Unnecessary parentheses

lib/util/fonts/resolveFontWeight.js
Munter: Yeah, this is pretty close to the described algorithm. But I think the description doesn't have the luxury of already having a limited subset of available weights, so it describes checking all weights.

I thought about early returns like `lastHalf[0] || firstHalf[firstHalf.length - 1]`, but it feel less clear from reading the code what the purpose it

lib/util/fonts/snapToAvailableFontProperties.js
Munter: I'm using `stretchStartIndex + 1` in multiple places as a single construct, so I opted to keep these to make it easier to recognise

lib/util/fonts/snapToAvailableFontProperties.js
Munter: Fixed

lib/util/fonts/snapToAvailableFontProperties.js
Munter: Fixed

papandreou:
Awesome! I'll take a look at CSS animations as soon as I get the chance.

PR #766, opened 2017-08-19
Automatically detach incoming relations when an asset is removed

papandreou:
This is a spinoff from #761, namely this idea:

* Disallow removing assets that have incoming relations -- leave them in the graph (maybe .unload() them instead). The benefit is that relations won't have a "dangling" state where they don't point to an instantiated asset, thus simplifying the model a lot. Peter: Consider just detaching all incoming relations when an asset is removed.

### DIFF
--- a/lib/AssetGraph.js
+++ b/lib/AssetGraph.js
@@ -223,49 +223,70 @@ class AssetGraph extends EventEmitter {
     }
 
     /**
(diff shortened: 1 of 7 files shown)


papandreou:
@Munter We discussed the idea of this a few months ago. Any thoughts about how it turned out?

PR #771, opened 2017-08-23
Implement JavaScriptImport and JavaScriptExport relations

papandreou:


papandreou:
> Isn't `export foo from 'bar.js'` just sugar syntax for `import foo from 'bar.js'; export foo`?

Yes, basically it is, except that you can also `export * from 'bar.js'` which I don't believe is expressible with `import`.

> I'm wondering if an export is even a relation

Good shout -- I agree when it comes to `export` without a url. This only intended to model `ExportNamedDeclaration` with a source and `ExportAllDeclaration` as relations. I notice now that it `ExportNamedDeclaration` doesn't always have a source, eg. `export function foo () {}`, and I just pushed 4b0f910 which skips all `Export*` nodes that don't explicitly reference a source.

It would certainly be possible to merge `JavaScriptImport` and `JavaScriptExport` together, but I felt it was clearer to name the `export ... from 'url'` variant something else since it's not an `import` statement. OTOH it's also weird that not every `export ...` leads to a `JavaScriptExport` relation being created.

Do you think we should merge them?

papandreou:
Although if we merge them, `javaScript.addRelation(...)` without a node would have to arbitrarily choose whether to add an `import ...` or an `export ...`.

papandreou:
I'll merge this in its current state. It'd be easy to turn `JavaScript{Im,Ex}port` into one class later if we decide to go for that.

PR #778, opened 2017-09-11
[bug] Add test-cases for spaces.

bramstein:
It seems the HTML minification step incorrectly handles various space characters. I've added a couple of test cases (that are failing), but the list is by no means exhaustive.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "estraverse": "^4.2.0",
     "gettemporaryfilepath": "^1.0.0",
     "glob": "^7.0.5",
-    "html-minifier": "^3.5.2",
+    "html-minifier": "^3.5.5",
     "imageinfo": "^1.0.4",
     "jsdom": "^11.2.0",
     "lodash": "^4.11.2",


papandreou:
Thanks! This could turn out to be a problem with [html-minifier](https://github.com/kangax/html-minifier) or [jsdom](https://github.com/tmpvar/jsdom), but we'd better track it down :)

papandreou:
Yup, it's html-minifier. I tried running the hair space test and logged `this._text.indexOf('\u200a')` before and after running `this._text` through html-minifier:

```
before 7
after -1
```

papandreou:
It's most likely a side effect of those space characters being included in `\s` in regular expressions:

```js
> /\s/.test('\u200a')
true
```

papandreou:
Let's see if https://github.com/kangax/html-minifier/pull/849 helps...

papandreou:
Fixed in 3.8.1 and assetgraph-builder 5.6.0.

bramstein:
Thanks! I can use my crazy spaces now 😜

PR #780, opened 2017-09-20
Trace CSS animations

papandreou:
- [x] Make sure that `font-weight` ranges are expanded correctly. Right now `getTextByFontProperties` will return entries with eg. `{'font-weight': '100-400'}`


papandreou:
This should be ready now.

PR #781, opened 2017-09-21
Font subsetting: Pick up intermediate font-weight values from CSS transitions to pseudo class states

papandreou:


Munter:
Looks good!

PR #782, opened 2017-09-21
transforms/subsetGoogleFonts: Added fontDisplay option

Munter:


lib/transforms/subsetGoogleFonts.js
```diff
@@ -373,6 +382,24 @@ module.exports = function (options) {
                         fontStyleBundle = new AssetGraph.Css(assetGraph.resolveAssetConfig({
                             text: source
                         }));
+
+                        if (fontDisplay) {
+                            fontStyleBundle.parseTree.nodes.forEach(function (atRule) {
+                                var declarations = atRule.nodes;
```
papandreou: Hmm, this won't detect `@font-face` rules nested inside `@media` queries.

Also, we should only dive into `@font-face` rules here in the code that follows? (`atRule.name === 'font-face'`)

Munter: Yeah, this is pretty native and will probably only work for Google fonts, where I can make a lot of assumptions about the stylesheet. I think the generic case might be a bit harder

papandreou: Fair :)

lib/transforms/subsetGoogleFonts.js
```diff
@@ -133,11 +133,20 @@ var fontFormatUA = {
...
+    'fallback',
+    'optional'
+];
+
 module.exports = function (options) {
     options = options || {};
     var subsetPerPage = !!options.subsetPerPage;
     var inlineSubsets = !!options.inlineSubsets;
     var inlineCss = !!options.inlineCss;
+    var fontDisplay = validFontDisplayValues.includes(options.fontDisplay) && options.fontDisplay;
```
papandreou: The build fails with node 4 due to the use of `Array#includes`

Munter: Fixed and rebased

PR #783, opened 2017-10-04
Feature/subset exact list items

papandreou:
This changes the font subsetting code to only include the characters needed to render the actually (potentially) exercised counter values. Also covers the hypothetical cases where eg. an element is only `display:list-item` for certain `@media`, and something else in other cases. This is accomplished by keeping track of all potential values for each counter and list settings. In all but the most obscure cases there will only be one possible value for each counter, so this change shouldn't cause too much overhead.

~~~Only covers `decimal` and custom `@counter-style` for now. Adding roman numerals and those other crazy built-in counter styles would be easy (copy/paste job from cldr), but it's probably not that important, and it would bloat the code. Let's wait a bit and consider releasing separate npm modules for the rendering of those.~~~ (had a couple of hours on a plane, so I went ahead and did this anyway, didn't bloat too much I think)

- [x] Clean up a bit, untangle the function that adjusts the potential counter values.


papandreou:
@Munter, I believe this is ready now. It also brings in some correctness fixes wrt. tracing non-inherited CSS properties.

I'm officially out of crazy ideas for refining the tracing ;)

Munter:
Outstanding. Hat tip to you sir! 

> I'm officially out of crazy ideas for refining the tracing ;)

Not to ruin your day, but... https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters

Munter:
Oops. Merged to master. I'll undo that a bit

papandreou:
> Not to ruin your day, but... https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Lists_and_Counters/Using_CSS_counters

This PR does support the full extent of CSS counters (notice that we're now tracing `counter-{set,reset,increment}` and doing housekeeping of all the possible values per counter. It also supports custom `@counter-style` declarations, a spec that's so far implemented in Firefox 8-)

papandreou:
Oh sorry, I must have targeted the wrong branch :)

Issue #767, opened 2017-08-19
v4 ideas


papandreou:
- [ ] Maybe get rid of the TransformQueue, just call each transform separately and `await`
- [ ] Allow transforms to fulfill with at custom value (don't force it to be the AssetGraph instance)
- [ ] Plugin API: .use(...) with a function that can install assets, relations, and named transforms
- [ ] Pluggable relations via asset-specific visitor functions
- [ ] Spike: Make `parseTree`, `text`, `rawSrc` async again

devmattrick:
I'm definitely interested in a Plugin API. I've been hoping to migrate a project over to something like this but I'm currently using a "nonstandard" templating engine. It would be great to be able to add support for it.

papandreou:
@devmattrick, thanks for weighing in. When we get around to this, we should certainly keep that use case in mind. Could you show an example of your template syntax and how you're loading/referring to the templates? It could be that it's already possible to get them to work.

devmattrick:
@papandreou I'm currently using [Nunjucks](https://mozilla.github.io/nunjucks/templating.html), which is a somewhat obscure templating engine, but its pretty darn similar to vanilla HTML.

papandreou:
@devmattrick Cool, are your nunjucks templates located directly in your main .html, or are they in separate files?

devmattrick:
@papandreou the template files are all separated. The entrypoints would be each Nunjucks file the views/ folder.

papandreou:
@devmattrick Can the nunchucks files distinguished from regular .html files based on a file extension or something similar?

... And I assume that what you want to accomplish is to discover the relations in the pure HTML parts of the nunchucks files?

devmattrick:
@papandreou I use `.njk` as a file extension for them, but I believe `.nunjucks` and even `.html` is sometimes used by other people. I would pretty much just like to resolve the images, stylesheets, and scripts in the HTML portions, the actual templating could probably be handled by the Nunjucks compiler itself, I think.

PR #785, opened 2017-10-18
Propagate the fulfillment/return value of the transform to the returned promise (or the callback)

papandreou:
As discussed on the gitter channel today.

It's one of the items here: https://github.com/assetgraph/assetgraph/issues/767

semver-major because the fulfillment value used to be the AssetGraph instance.


papandreou:
:hugs:

PR #786, opened 2017-10-27
Subset google fonts using font face declaration

Munter:
This eliminates a lot of assumptions that were specific to Google fonts.

- Fixes error where including google font id's that resolved to no actual font, would propagate down to the tracing code as an actual available font. New behavior downloads google font stylesheet and traces `@font-face`-declarations instead of assuming resonses based on the URL
- Paves the way for local font subsetting


papandreou:
Haven't done a thorough review, but we've talked about this a lot, and it would be really great to get it in :+1:

I'm a little worried about the conflict we'll get when merging it to master -- but let's cross that bridge when we get there.

PR #787, opened 2017-11-16
Model <noscript>...</noscript> as an inline Html asset

papandreou:


Munter:
Yay!

Munter:
This looks great

PR #788, opened 2017-11-19
Support conditional comments when tracing subsets

papandreou:
Bonus feature: Support nested `@media` queries.


Munter:
Great stuff!

PR #790, opened 2017-11-20
Feature/subset noscript

papandreou:
Prerequisite: https://github.com/assetgraph/assetgraph/pull/789

Tested against a crude backport of https://github.com/assetgraph/assetgraph/pull/787 on v3.

The tests can be unskipped once this has been merged to master or we land a backport of HtmlNoscript on v3.

### DIFF
--- a/lib/util/fonts/getTextByFontProperties.js
+++ b/lib/util/fonts/getTextByFontProperties.js
@@ -690,21 +690,8 @@ function getTextByFontProperties(htmlAsset, memoizedGetCssRulesByProperty) {
                     if (relation.type === 'HtmlNoscript' && relation.node === node) {
                         noscriptStack.push(true);
                         var noscriptDocument = relation.to.parseTree;
-                        var isWithinBody = false;
                         for (var i = 0 ; i < noscriptDocument.childNodes.length ; i += 1) {
                             var childNode = noscriptDocument.childNodes[i];
-                            // Don't proceed unless we're between
-                            // <!--ASSETGRAPH DOCUMENT START MARKER--> and <!--ASSETGRAPH DOCUMENT END MARKER-->
-                            if (childNode.nodeType === childNode.COMMENT_NODE) {
-                                if (childNode.nodeValue === 'ASSETGRAPH DOCUMENT START MARKER') {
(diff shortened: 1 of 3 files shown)


papandreou:
Rebased now that https://github.com/assetgraph/assetgraph/pull/789 went in.

lib/util/fonts/getTextByFontProperties.js
```diff
@@ -684,8 +685,42 @@ function getTextByFontProperties(htmlAsset, memoizedGetCssRulesByProperty) {
...
 
-            // Inputs might have visual text, but don't have childNodes
-            if (node.tagName === 'INPUT' && visualValueInputTypes.indexOf(node.type || 'text') !== -1) {
+            if (node.tagName === 'NOSCRIPT') {
+                htmlAsset.outgoingRelations.some(function (relation) {
+                    if (relation.type === 'HtmlNoscript' && relation.node === node) {
+                        noscriptStack.push(true);
+                        var noscriptDocument = relation.to.parseTree;
+                        var isWithinBody = false;
+                        for (var i = 0 ; i < noscriptDocument.childNodes.length ; i += 1) {
```
Munter: This next block seems like a lot of gynmastics to work against our document markers. Not a nice user experience to duplicate any time you need to traverse a noscript tag. Could we somehow encapsulate this in the parseTree getter?

papandreou: Yeah, you're right. I think this boilerplate stems from the fact that I copy/pasted from HtmlConditionalComment, which needs to be able to contain unbalanced HTML, `<html>` and other weirdness. I'll see if it can be simplified :+1:

papandreou: Done (+ updated https://github.com/assetgraph/assetgraph/pull/787), take another look now :)

papandreou:
Assuming this is OK now :)

Munter:
Yeah. I'm backporting the noscript relation to v3. Should land any minute now

papandreou:
Ah, cool, I also did that on a branch that I only have locally, but it's fairly easy :)

PR #793, opened 2017-11-23
 Make Asset#type a getter that offers a guess based on Content-Type, incoming relations, file extension etc.

papandreou:
As discussed on the gitter channel. Was easier than I expected.

I've copy/pasted some code from Asset#loadAsync, and there's still a chance to dedupe it some. LMK if it seems worthwhile to proceed :)

### DIFF
--- a/lib/assets/Asset.js
+++ b/lib/assets/Asset.js
@@ -269,7 +269,7 @@ class Asset extends EventEmitter {
                 }
                 if (commonTargetType) {
                     if (Class && !AssetGraph[commonTargetType].prototype['is' + Class.prototype.type]) {
-                        this.from.warn(`${this.urlOrDescription} used as both ${Class.prototype.type} and ${commonTargetType}`);
+                        this.assetGraph.warn(`${this.urlOrDescription} used as both ${Class.prototype.type} and ${commonTargetType}`);
                     } else {
                         Class = AssetGraph[commonTargetType];
                     }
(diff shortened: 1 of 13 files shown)


lib/assets/Asset.js
```diff
@@ -114,6 +114,39 @@ class Asset extends EventEmitter {
...
+                return typeFromExtension;
+            } else if (commonTargetType) {
+                return commonTargetType;
+            }
+        }
+        return this._type;
+    }
+
+    set type(type) {
+        this._type = type;
```
Munter: Should there be a setter at all, and if so, should this not include the logic to detect sub typing or switches to different asset type family branches?

test/transforms/populate.js
```diff
@@ -14,7 +14,8 @@ describe('transforms/populate', function () {
                 followRelations: {type: query.not('HtmlStyle')}
             });
 
-        expect(assetGraph, 'to contain no assets', 'Css');
+        expect(assetGraph, 'to contain asset', 'Css');
+        expect(assetGraph, 'to contain asset', { type: 'Css', isLoaded: false });
```
Munter: This seems to just be a more specific version of the previous assertion. Kill the less specific version?

Munter:
It's great to see this improving the quality of the graph from your test adjustments :D

test/transforms/populate.js
papandreou: You're right, fixed! It was late :)

lib/assets/Asset.js
papandreou: I assumed that there would be some places where the `type` property would be set by `createAsset` or friends, but apparently I didn't bother to check. It works fine without the setter, and I've removed it now.

PR #794, opened 2017-11-25
Stop using type:'Asset' to mean unknown type in the public API, use undefined instead

papandreou:


Munter:
I like it :)

papandreou:
Let's pull it in, then :)

PR #795, opened 2017-11-25
Update dependencies

papandreou:


papandreou:
Ah, it's also updating some dependencies needlessly. I guess I should have run `npm update` first.

PR #796, opened 2017-11-26
Also trace Firefox' default stylesheet when subsetting fonts

papandreou:
Not entirely happy with the state of this, mainly due to the manual expansions I had to do in the CSS.


papandreou:
Apparently those `:-moz-any(...)` constructs really do cause selectors to have different specificities depending on what they match: https://github.com/keeganstreet/specificity/issues/1#issuecomment-346997589 -- bleh!

I think that means we can't do much better than the explosion, except for getting it automated somehow.

papandreou:
Another downside of this is that the Firefox stylesheet adds a metric shitton of entries in the output of `getCssRulesByProperty`:

The Chromium default stylesheet (number of entries by CSS property):

font-family 8
font-style 6
font-weight 11
content 4
quotes 0
list-style-type 10
display 96
animation-name 0
transition-property 0
transition-duration 0
counter-increment 0
counter-reset 0
counter-set 0
TOTAL: 135

The Firefox default stylesheet (number of entries by CSS property):

font-family 8
font-style 6
font-weight 1373
content 8
quotes 0
list-style-type 66
display 1451
animation-name 0
transition-property 0
transition-duration 0
counter-increment 0
counter-reset 0
counter-set 0
TOTAL: 2912

This slows down the `getTextByFontProperties` tests from about 3 seconds to 30-something seconds.

papandreou:
Hand-editing the default stylesheet helped a bit.

Munter:
I'm slowly beginning to worry about the runtime. I'm seeing 5 seconds just for tracing already just on my blog with almost no content. This amount of selectors seems a bit wild

papandreou:
> I'm slowly beginning to worry about the runtime

Hehe, yeah, me too, except I've been concerned about it for some time :). Maybe we should try to cut down the Firefox stylesheet even further so it only accounts for those actual differences we've identified.

papandreou:
> I'm seeing 5 seconds just for tracing already just on my blog with almost no content

Is that with or without this change?

Munter:
That's with the current v3. Chrome browser defaults only

papandreou:
Yeah, it's a real concern. I think https://github.com/assetgraph/assetgraph/commit/9c6d628ab9e8bc575d433991925ecd4c4d72f927 might have helped in the general case, though -- could you try to run the same case against current v3?

Munter:
Rebased on v3.

Current time to trace mntr.dk (12 pages): 8 seconds :(

But I guess we need it in order to avoid those pesky original font loads

papandreou:
Thanks for checking again. What's the difference between the runtime with and without this change?

Munter:
Subfont v1.x: `4.127 secs: collectTextsByPage`
Subfont vNext: `4.499 secs: collectTextsByPage`
Subfont vNext + Firefox: `7.736 secs: collectTextsByPage`

papandreou:
Okay, thanks for doing that. We'll definitely need to optimize something before we can consider merging this.

Munter:
https://ag-traceprop-flame.surge.sh

papandreou:
I boiled the Firefox default stylesheet down to that `b, strong { font-weight: bolder }` rule now, and it seems like the performance was regained. We can reintroduce more parts as we identify differences that actually matter.

Munter:
So if I understand this correctly, now the firefox stylesheet is depending on the chromium stylesheet to provide all the other defaults that are identical?

papandreou:
@Munter, yeah, correct. I've "layered" the single remaining exception for Firefox is "layered" over the Chromium default stylesheet by putting it after the Chromium stylesheet in the array so it has a higher specificity tiebreaker, and the fact that it has a predicate makes `traceProp` also include the possibility of it not matching whenever it hits it.

Munter:
Sounds good. As long as we remember this dependency if we ever need to replace the chromium default

papandreou:
It's perfectly safe to replace/upgrade the Chromium default.

Munter:
Released in v3.13.0: https://github.com/assetgraph/assetgraph/releases/tag/v3.13.0

PR #798, opened 2017-12-03
subsetFonts: Don't break when there's an existing <link rel=preload as=font ...>

papandreou:
Also, don't inject another one :)


lib/transforms/subsetFonts.js
```diff
@@ -567,15 +567,21 @@ module.exports = function (options) {
...
-                                    as: 'font'
-                                });
-                            });
 
-                            preloadRelations.forEach(function (rel) {
-                                rel.attach(htmlAsset, 'before', insertionPoint);
+                                var preloadHint = assetGraph.findRelations({
+                                    type: 'HtmlPreloadLink',
+                                    as: 'font',
+                                    to: fontAsset
```
Munter: This will find any preload relation to the font, including ones on different pages. We need to match the from Html asset as well

PR #807, opened 2017-12-18
Asset: Implement a Proxy-based query getter/setter

papandreou:


Munter:
This will make things a lot easier

PR #810, opened 2017-12-23
Update to cssnano@next, replace Css#minify with a minifyCss transform

### LINKED ISSUE #809: cssnano/postcss is outputting deprecation notice to the console
This usually wouldn't be much of an issue except when using assetgraph with a CLI client it makes the output ugly. The following is output to the console

```
Node#moveTo was deprecated. Use Container#append.
```

Issue is detailed here: https://github.com/ben-eb/cssnano/issues/379

Looks like the solution is to either downgrade postcss or move to `cssnano@next`. Any thoughts? If it's relatively straightforward then I'm happy to PR.

papandreou:
Fixes #809


lib/transforms/minifyCss.js
```diff
@@ -0,0 +1,33 @@
...
+        }
+        for (const cssAsset of assetGraph.findAssets(queryObj)) {
+            const result = await cssnano.process(cssAsset.parseTree, undefined, {
+                preset: [
+                    'default',
+                    {
+                        svgo: false,
+                        safe: true,
+                        normalizeUrl: false, // Buggy
+                        mergeLonghand: false,
```
Munter: Longhand merging seems useful. Does enabling it cause trouble somehow?

papandreou: There used to be a bug that caused it to kick in even when all the longhand properties weren’t defined, so it didn’t preserve the semantics, yet was classified as a “safe” transform. It caused problems in a project of mine where some styling got mangled.

Sounds like [it’s fixed now](https://github.com/ben-eb/cssnano/issues/358#issuecomment-283696193). We can certainly try re-enabling it, but Ididn’t want to mix that into this pr.

davej:
Thanks for this! Should the `isMinified` property of the minified Css assets be set to true after using this transform?

🎄

davej:
Do you think it's worth adding minification transforms of other asset types for consistency?

papandreou:
@davej, hmm, good q, the way I see it, `isMinified` is mostly an interval property that decides whether the asset is to be minified/compacted "on the way out" when it's eventually serialized. That's no longer necessary for Css now that it's an explicit action that mutates the AST. It's still necessary for JavaScript and Html, though.

... But for consistency I guess we should either reintroduce `Css#isMinified` or explicitly mark `JavaScript#isMinified` and `Html#isMinified` as "private.

davej:
Hmm, yeah that's a bit confusing since `isLoaded`, `isDirty` etc… describe the current state of the asset. I have been using `isMinified` in queries. Perhaps a more accurate name would clear up the confusion `shouldMinify`, `toBeMinified`.

papandreou:
@davej, that sounds like a really good idea. I like `toBeMinified`.

lib/transforms/minifyCss.js
```diff
@@ -0,0 +1,33 @@
...
+
+module.exports = queryObj => {
+    return async assetGraph => {
+        if (queryObj) {
+            queryObj = assetGraph.query.and(queryObj, defaultQueryObj);
+        } else {
+            queryObj = defaultQueryObj;
+        }
+        for (const cssAsset of assetGraph.findAssets(queryObj)) {
+            const result = await cssnano.process(cssAsset.parseTree, undefined, {
```
davej: @papandreou: Should `cssnano.process` be wrapped in a try/catch block with errors being passed to `assetgraph.warn`?

papandreou: @davej, hmm, I guess that can't hurt? Haven't ever seen it return a rejected promise, though.

papandreou:
https://github.com/assetgraph/assetgraph/commit/ea7c1e4327a908bfe77f365f19415fa4020d3eb1

lib/transforms/minifyCss.js
papandreou: https://github.com/assetgraph/assetgraph/commit/ea7c1e4327a908bfe77f365f19415fa4020d3eb1

davej: @papandreou Hmm, does it not reject if the input css is malformed or non-css content.

papandreou: Ah, you're right, although that would come from the execution of `cssAsset.parseTree` if the asset hasn't already been parsed. I've added a test of that scenario now: https://github.com/assetgraph/assetgraph/commit/12afea957f364e8b02997b704c302f5db0ae2b05

PR #812, opened 2017-12-27
Clean up the cssnano config, give normalizeUrl and mergeLonghand another chance

papandreou:
Following up from #810.

### DIFF
--- a/lib/transforms/minifyCss.js
+++ b/lib/transforms/minifyCss.js
@@ -15,10 +15,6 @@ module.exports = queryObj => {
                     'default',
                     {
                         svgo: false,
-                        safe: true,
-                        normalizeUrl: false, // Buggy
-                        mergeLonghand: false,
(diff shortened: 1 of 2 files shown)


lib/transforms/minifyCss.js
```diff
@@ -15,10 +15,6 @@ module.exports = queryObj => {
                     'default',
                     {
                         svgo: false,
-                        safe: true,
-                        normalizeUrl: false, // Buggy
-                        mergeLonghand: false,
-                        autoprefixer: false,
```
Munter: Does this removal mean autoprefixer will be run?

papandreou: No, autoprefixer is not included in the `default` preset: https://github.com/ben-eb/cssnano/blob/af7c3f4f8a31b9356cf69daf88cd7da56357b633/packages/cssnano-preset-default/src/index.js

PR #814, opened 2017-12-27
Replace the built-in query engine with sift.js

papandreou:
This is an attempt to make AssetGraph more approachable by changing to a mongo-like query syntax provided by an external library (https://github.com/crcn/sift.js).

Haven't made up my mind about it yet -- LMKWYT :)

CC: @davej


davej:
In general, I like the idea of shipping off non-core features like querying to a third-party lib so +1.

No strong feelings either way about the syntax. I guess it's good since some people may already be familiar with Mongo and if they're not then the Mongo Query docs seem pretty comprehensive. 👍

Munter:
As discussed in person I agree with pulling in sift to replace our custom query engine. Given some extra sugar on url accessors I don't even think we need to retain the ability to keep our regex literal query syntax and can live with the `{$regex: ... }` thing that sift uses

papandreou:
Rebased on master.

Here's what it looks like if we take advantage of the features introduced in #822: 
https://github.com/assetgraph/assetgraph/compare/feature/siftAndNewGetters

(if we want to go for it, let's merge #822 first, then cherry-pick the commits after the merge on https://github.com/assetgraph/assetgraph/commits/feature/siftAndNewGetters)

PR #815, opened 2017-12-27
isMinified => _toBeMinified

papandreou:
As suggested here: https://github.com/assetgraph/assetgraph/pull/810#issuecomment-353972716


papandreou:
It's an internal attribute, so no tests were touching it. Come to think of it, maybe we should name it `_toBeMinified`.

papandreou:
Done.

PR #816, opened 2017-12-28
Trace ::first-letter and ::first-line when subsetting fonts

papandreou:
This is a bit of a mess, I know. Tracing those required a restructuring of how text and pseudo elements are traced.

Bonus feature: `::placeholder`

### DIFF
--- a/lib/util/arePredicatesExhaustive.js
+++ b/lib/util/arePredicatesExhaustive.js
@@ -1,0 +1,21 @@
+var expandPermutations = require('./expandPermutations');
+
+function arePredicatesExhaustive(predicatesArray) {
+    var seenPredicateNames = {};
+    predicatesArray.forEach(function (predicates) {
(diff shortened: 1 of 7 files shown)


papandreou:
The reordering happened as a result of tracing the node, pseudo elements, and children in a slightly different order. It's a bit intricate because of the counter adjustments etc. I was careful to check that the same entries were there in all the affected tests.

Might be possible to get it back to the previous state with a bit of work.

Munter:
The order doesn't matter in our later processing, so I don't care much. I was just wondering :)

Munter:
Now we just need to figure out how to port this to v4 :)

papandreou:
:+1:

It should merge cleanly to master. I've deliberately not touched `{test,lib}/util/fonts/getTextByFontProperties.js` there because this was still open.

PR #823, opened 2018-01-05
Test case ported from assetgraph-builder

gustavnikolaj:
Test case ported from: https://github.com/assetgraph/assetgraph-builder/pull/545

Can be removed and closed - there's already another test case testing mostly the same thing, it's just added here to proof that the assetgraph-builder PR is necessary as the error is triggered by a later transform in buildProduction and not straight out of bundleWebpack.


papandreou:
Fixed the problem in assetgraph-builder now, released in 5.7.4.

PR #828, opened 2018-01-22
reviewContentSecurityPolicy: Omit the schema when adding new source expressions to CSP directives

### LINKED ISSUE #827: reviewCsp: See if we can squeeze out some bytes by omitting the scheme when we know the page is served via https
https://scotthelme.co.uk/optimising-twitters-csp-header/

papandreou:
As recommended by Scott Helme in https://scotthelme.co.uk/optimising-twitters-csp-header/

@gustavnikolaj, would be great to test this with a real application if you have some time :)

Fixes #827

### DIFF
--- a/lib/transforms/reviewContentSecurityPolicy.js
+++ b/lib/transforms/reviewContentSecurityPolicy.js
@@ -53,27 +53,18 @@ module.exports = function (queryObj, options) {
             });
         }
 
-        function originFromUrl(url, directive, isInProtocolRelativeContext) {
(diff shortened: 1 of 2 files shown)


Munter:
Merging. I can live with the coverage drop and the appveyor tests haven't been set up properly yet anyway

papandreou:
Thanks! The coverage drop is due to the removal of well-tested lines, there's no change in the number of uncovered lines (which would be a better metric).

PR #832, opened 2018-03-04
Drop the minifyCss transform and go back to a Css#minify method

papandreou:
Async, though, because of the new cssnano API

As discussed on gitter. The `minifyCss` transform was just a bad idea.

// CC: @davej


davej:
Awesome. Thanks! 👍

PR #851, opened 2018-04-21
Filter out assets that are essentially file redirects when writing to disc

Munter:


lib/transforms/writeAssetsToDisc.js
```diff
@@ -25,7 +25,15 @@ module.exports = (queryObj, outRoot, root) => {
...
   return async function writeAssetsToDisc(assetGraph) {
     await Promise.map(
-      assetGraph.findAssets(Object.assign({ isInline: false }, queryObj || {})),
+      assetGraph.findAssets(
+        Object.assign(
+          {
+            isInline: false,
+            fileRedirectTargetUrl: undefined // FIXME: Dependency on non-public api. Smelly
+          },
+          queryObj || {}
```
papandreou: `|| {}` isn’t necessary.

PR #864, opened 2018-05-21
Add missing HtmlResourceHint valid as-property values according to up…

Munter:
…dated spec

Fixes https://github.com/Munter/subfont/issues/25

### DIFF
--- a/lib/relations/HtmlPrefetchLink.js
+++ b/lib/relations/HtmlPrefetchLink.js
@@ -16,7 +16,8 @@ class HtmlPrefetchLink extends HtmlResourceHint {
 
     if (this.as) {
       this.node.setAttribute('as', this.as);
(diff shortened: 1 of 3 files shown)


Munter:
I'm not quite sure how this warning could cause subfont to not do it's work. Are we swallowing an exception somewhere?

papandreou:
From the output quoted in https://github.com/Munter/subfont/issues/25 it seems like it did work, it just spewed out those confusing warnings?

KyleAMathews:
We didn't investigate deeper why but can confirm that subfont didn't do any rewriting of the HTML until we'd removed the `<link preload as="fetch">`

KyleAMathews:
Thanks for the quick fix! Launching a new website with this soon! :-D

PR #866, opened 2018-05-26
subsetFonts: Support the 'font' shorthand

papandreou:


papandreou:
@Munter, looks right to me, thanks! Merged.

PR #869, opened 2018-06-01
Add more verbose errors for missing glyphs to allow the user to locat…

Munter:
…e the glyph on theor page

Switched to using `opentype` for inspecting the resulting subsets CMAP and inspect the available glyphs from that. Under the assumption that if a known used glyph should be part of the subset, but isn't included, we can infer that the glyph is also missing in the original font.

With that change we can avoid running `pyftsubset` in a way where it throws and reruns in a more forgiving way. We will also get missing glyph warnings when subsetting using google fonts!


Munter:
I don't understand this linting error. On my machine the linter passes with no problems

papandreou:
The lint errors came about because of prettier 0.13, which included these changes: https://prettier.io/blog/2018/05/27/1.13.0.html#javascript

package.json
```diff
@@ -42,10 +42,12 @@
     "html-minifier": "^3.5.8",
     "imageinfo": "^1.0.4",
     "jsdom": "^11.11.0",
+    "lines-and-columns": "^1.1.6",
     "lodash": "^4.11.2",
     "memoizesync": "1.1.1",
     "mkdirp": "^0.5.1",
     "normalizeurl": "^1.0.0",
+    "opentype": "github:bramstein/opentype",
```
papandreou: Can we get on an officially released version? Do we need to poke Bram?

test/transforms/subsetFonts.js
```diff
@@ -2771,9 +2773,9 @@ describe('transforms/subsetFonts', function() {
           expect(warnSpy, 'to have calls satisfying', function() {
             warnSpy({
               message: expect
-                .it('to contain', 'OpenSans.ttf is missing these characters')
-                .and('to contain', 'U+4E2D (中)')
-                .and('to contain', 'U+56FD (国)')
+                .it('to contain', 'Missing glyph fallback detected')
+                .and('to contain', '\\u{4e2d} (中)')
+                .and('to contain', '\\u{56fd} (国)')
```
papandreou: Why is the `\u{nnnn}` syntax better?

Munter: Because Mathias says so

img_20180601_210131

papandreou: Weak :)

Munter: I know. But the timing was perfect :)

This is the article where he outlines the unicode escape sequence problems: https://mathiasbynens.be/notes/javascript-unicode#escape-sequences

This syntax lets you describe astral planes as well instead of having to create surrogate planes

papandreou: Okay, I'm buying that :)

Next stop --- the stars 🌃

papandreou:
After we merge this, I think we should keep prettier on a tighter leash, maybe we can set it to `~0.13`? I'm actually wondering why depfu didn't notify us of the breakage before.

package.json
bramstein: What is it doing? You might be better of using something like fontkit.

Munter: All I need is a tiny part of the cmap information. Namely which codepoints are present in the font file. I couldn't find that in the other libraries I tried. At least not for woff/woff2

Munter: This package would be the exact thing I needed if it supported a wider range of font formats. https://www.npmjs.com/package/font-cmap

Munter: ah, it does look like I can get that with `fontkit.openSync(tempWrite.sync(fontUsage.subsets.woff)).characterSet`. Their docs do require quite specific reading through. No good keywords to search on

Munter:
I replaced opentype with fontkit. I think this is good to go

bramstein:
Maybe add a license file for the fonts?

PR #880, opened 2018-06-24
es6ify font code

papandreou:


papandreou:
Already landed manually on master.

PR #899, opened 2018-07-14
subsetFonts: Gather stats about the byte sizes of the original and subset fonts

papandreou:


papandreou:
This should be ready now.

papandreou:
Released in 4.12.0

PR #904, opened 2018-07-16
Added function to download google fonts

Munter:
I opted for a setup where I create a google font id for each passed in `fontProps` object, but use the google response CSS to populate the resulting CSS properties. This is to ensure that the response keeps in sync with what google interprets the fontProps as.

I added unicode-range and font-display setting ability right away, and also a subsetting ability.

Even though were going to have a few stringifications and parsings of the fonts I think it's worth it to keep the function agnostic. We can make a transform that takes the resulting CSS font-face declaration stylesheets, concats them and extracts the fonts to a given URL later.

### DIFF
--- a/lib/util/downloadGoogleFonts.js
+++ b/lib/util/downloadGoogleFonts.js
@@ -103,11 +103,11 @@ async function downloadGoogleFonts(
   return results.map(result => {
     const declarationStrings = [];
 
-    Object.entries(result).map(([property, value]) => {
+    for (const [property, value] of Object.entries(result)) {
       if (property !== 'src') {
         declarationStrings.push(`  ${property}: ${value};`);
       }
-    });
+    }
 
     const sources = result.src.map(([font, format]) => {
       return `url('${font.dataUrl}') format('${format}')`;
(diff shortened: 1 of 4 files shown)


lib/util/downloadGoogleFonts.js
```diff
@@ -0,0 +1,120 @@
...
+      }
+
+      return fontProps;
+    })
+  );
+
+  return results.map(result => {
+    const declarationStrings = [];
+
+    Object.entries(result).map(([property, value]) => {
```
papandreou: Could you use a `for ... of` here? It's a bit weird to use `Array#map` and discard the return value.

Munter:
Since I went for downloading a separate CSS per font variant for simplicity, I think it might be better to change the API from taking an array in and returning an array to just taking a single font props object and return the Css string. If the user wants more variants at the same time they can just make more calls. There's no difference in the network usage

Munter:
The original idea was that I would use the format as the top level iteration and request all the font variants in the same request for the CSS. But I ended up deciding that having to map the responses multiple font-face blocks back to the input font properties array was needlessly complex. So when I switched it around to have the font variant as the top level iteration I just forgot that the array input/output don't really make a lot of sense anymore

PR #910, opened 2018-07-21
Fix multiple @font-face declarations with the same src but different font-family

papandreou:
This test is a stripped down version of a (weird, admittedly) construct I saw on https://v-fonts.com/:

```css
@font-face {
	src: url('/assets/fonts/dTJakobVariableConceptGX.woff2');
	font-family:'dT Jakob Variable';
	font-style: normal;
}
@font-face {
	src: url('/assets/fonts/dTJakobVariableConceptGX.woff2');
	font-family:'dT Jakob Variable (with OpenType standard ligatures enabled)';
	font-style: normal;
}
@font-face {
	src: url('/assets/fonts/dTJakobVariableConceptGX.woff2');
	font-family:'dT Jakob Variable (with OpenType discretionary ligatures enabled)';
	font-style: normal;
}
@font-face {
	src: url('/assets/fonts/dTJakobVariableConceptGX.woff2');
	font-family:'dT Jakob Variable (with OpenType contextual alternates enabled)';
	font-style: normal;
}
```

Which is then used like this:

```html
<li class="font">
  | <figure>
  | <figcaption>dT Jakob Variable (with OpenType discretionary ligatures enabled)</figcaption>
  | <div class="sample M L all-caps grayscale" style="font-family: 'dT Jakob Variable (with OpenType discretionary ligatures enabled)', 'Adobe Blank';" contenteditable="true" autocomplete="off" autocorrect="off" autocapitalize="off" spellcheck="false">
  | <div style="font-feature-settings: 'liga' 0, 'dlig' 1, 'calt' 0";>ABCDEFGHIJKLMNOPQRSTUVWXYZ <span>0123456789</span></div>
  | </div>
  |  
  | </figure>
  | </li>
```

We trace and snap this correctly (as separate font prop combos), but then it seems like we group both texts into the same bucket in the `fontInfo` so that only one the has a subset created. And we fail to patch up the `font-family` properties pointing at the other `@font-face`s.

I guess we would be able to create a common subset font for the two `@font-face` declarations in this case. ~~~However, this also happens when the `@font-face` declarations have different `font-style` or `font-weight` values, so I think it would be much easier to fix it by grouping by font url+`font-family` instead of only the font url when determining the subsets to create. I haven't actually looked into the code, though.~~~


papandreou:
After thinking about it for a bit I'm fairly sure that what I wrote before is just wrong -- creating a common subset of the font is fine, even though it's being referenced in multiple `@font-face` declarations that assign different `font-family` names. I've changed to implementation so it keeps track of all `font-family` values that are in play for each font, then adds all of the to the `webfontNameMap` so the subset gets injected. I think this will work.

PR #923, opened 2018-08-23
Use new docdash features

Munter:


papandreou:
Awesome, looks great! Especially the search feature.

Munter:
I'm not so impressed with the collapsing. I read the documentation as if the menu would expand children if the selected section was active. Might require a bit more fiddling

PR #940, opened 2018-10-02
Deduplicate CSS and Fonts generated by transforms/subsetFonts

Munter:
Before:

image

After:

image


lib/transforms/subsetFonts.js
```diff
@@ -921,54 +919,65 @@ These glyphs are used on your site, but they don't exist in the font you applied
...
         }
+      }
+
+      const cssFileName = `fonts-${cssAsset.md5Hex.slice(0, 10)}.css`;
+      const cssUrl = subsetUrl + cssFileName;
+      const [identicalAsset] = assetGraph.findAssets({ url: cssUrl });
 
-        const cssFileName = `fonts-${cssAsset.md5Hex.slice(0, 10)}.css`;
-        cssAsset.url = subsetUrl + cssFileName;
+      if (identicalAsset) {
```
papandreou: Would it be possible to detect this condition earlier so we could avoid creating and adding the duplicate to the graph?

Munter: I think the answer is yes. But the current strategy would need a big revision. The part that gets the relevant stylesheet currently gets a string with base 64 inlined fonts. There was an attempt to dedupe on the per-page stylesheet level, but that failed and also seems like the wrong place to ensure this

Munter: I also need a test to verify. I thought a current one already covered this, but alas. Need to figure out how to trigger this condition

papandreou: Okay, cool. As long as a test is included, it will help us not reintroduce the bug when we get around to doing that revision :)

papandreou:
This would have to be redone against https://github.com/Munter/subfont (but I think it has been fixed in the meantime)

PR #950, opened 2018-11-15
downloadGoogleFonts: Fix retrieval of .ttf fonts

papandreou:
Sending an empty User-Agent causes GWF to deliver a ttf font without hints stripped.

The previous User-Agent was a mac one (so hints would be stripped) and actually caused GWF to send a .woff file.


Munter:
Nice catch

PR #951, opened 2018-11-22
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these LGTM [code quality badges](https://lgtm.com/projects/g/assetgraph/assetgraph/ci/#badges) to your project. They indicate how you care about code quality and encourage your future contributors to do the same.
To get an idea of the analyses reflected by these grades, check the [alerts discovered by LGTM](https://lgtm.com/projects/g/assetgraph/assetgraph).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,8 @@
 [![Build Status](https://travis-ci.org/assetgraph/assetgraph.svg?branch=master)](https://travis-ci.org/assetgraph/assetgraph)
 [![Coverage Status](https://img.shields.io/coveralls/assetgraph/assetgraph.svg)](https://coveralls.io/r/assetgraph/assetgraph?branch=master)
 [![Dependency Status](https://david-dm.org/assetgraph/assetgraph.svg)](https://david-dm.org/assetgraph/assetgraph)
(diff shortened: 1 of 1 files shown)


Munter:
Literally all the reported errors and warnings from that tool are about regular expressions that are tested to work. Also we have a linter telling us of we need some of those up. Not really useful feedback.

The suggestions section suggest things we explicitly disable the linter for because we want to achieve a side effect, as also commented on the very code blocks it highlights.

I'll pass on a tool that marks or code as worse quality than it really is

xcorail:
Hi @Munter 
Thanks for giving me this helpful feedback.

Indeed, the first ones on regular expressions seem to be false positive, thanks for reporting these.
Also, other users suggested that our tool takes into account the `eslint disable` comments, so that they don't have to add our `//LGTM` comment in addition ... our product team is discussing this internally atm, so I will use your feedback.

Thanks again and best of luck for assetgraph

PR #953, opened 2018-12-07
Add support for some CSS Fonts Module 4 features + variable font with non-variable fallback

papandreou:
Original idea: https://gitter.im/assetgraph/assetgraph?at=5bdd8c63a9c55b25749ebb3d

Adds support for:

* Subsetting all fonts mentioned in `font-family: foo, bar` that have `@font-face` declarations instead of only the first one. This is useful for cases where the highest prioritized one is a variable font followed by a non-variable fallback.
* `font-weight` ranges in `@font-face` declarations for variable fonts
* `font-stretch` ranges in `@font-face` declarations for variable fonts

Also includes these semi-related changes:

* Instead of normalizing the `font-` properties of each `@font-face` declaration early, we preserve the original values and only normalize internally in `snapToAvailableFontProperties` as part of finding the matching `@font-face` for a given trace. I made that change because CSS Fonts Module 4 enhances the `@font-face` properties to allow ranges, eg. `font-weight: 100 400`, which is disallowed elsewhere and thus got lost when if we apply the existing normalization.
  This has the side effect that the properties are not normalized in the returned "info" object, so we'll have to take that into account in the console reporting that `subfont` does.
* The introduction of a proper parser for `font-family` values.

Out of scope for this PR:

* `font-style: oblique <angle>`, it's not fully specified yet (https://www.w3.org/TR/css-fonts-4/#font-style-prop)
* For CSS animations and transitions, `font-weight` values are still traced in increments of 100, which can theoretically lead to wrong results if multiple variable fonts provide ranges with endpoints that are not multiples of 100. Probably not worth fixing.

TODO:

- [x] Support `font-weight` values that aren't multiples of 100: https://github.com/bramstein/css-font-parser/pull/7 (fixed for now by switching to the css-font-parser-papandreou fork)
- [x] Find a strategy that supports font-weight ranges here: https://github.com/assetgraph/assetgraph/blob/8e074f2c9e9fa338cea5d0a2d0109a037bc5476c/lib/transforms/subsetFonts.js#L1049-L1063


papandreou:
@Munter, fixed the last TODO now. I think this would be a nice step towards variable font support

PR #963, opened 2019-01-15
Handle non-ascii characters in urls

papandreou:
Browsers tolerate it, so let's try to handle them the same way.

Avoids the `[ERR_UNESCAPED_CHARACTERS]: Request path contains unescaped character` error that caused confusion in https://github.com/webpack/webpack.js.org/pull/2759


Munter:
Looks great!

papandreou:
I also tested non-ascii characters in iso-8859-1-encoded HTML documents in Chrome and Firefox. They're also sent as percent-encoded utf-8 octets, so I think this covers it.

PR #965, opened 2019-01-17
Jsdocto .d.ts

Munter:
This is an attempt at automatically generating typescript definitions from the existing jsdoc comments.

It seems like `types.d.ts` gets output correctly, but I haven't yet succeeded in actually getting vscode to pick the definitions up while doing local development or linking projects across the file system.

Example output: https://gist.github.com/Munter/06b6eb424867f2e333f20d93550fb3c7

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "specificity": "^0.4.0",
     "sw-precache": "^5.2.0",
     "teepee": "^2.31.1",
+    "tsd-jsdoc": "^2.1.2",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -61,6 +61,7 @@
     "specificity": "^0.4.0",
     "sw-precache": "^5.2.0",
     "teepee": "^2.31.1",
+    "tsd-jsdoc": "^2.1.2",
```
papandreou: I think this should be a dev dep?

PR #973, opened 2019-02-01
Add support for terser and use it as the default JavaScript compressor

papandreou:


papandreou:
@Munter, LGTY now?

Munter:
Yeah. Looks ready to merge

PR #976, opened 2019-02-15
Allow custom sw-precache configuration

salomvary:
As [discussed on the Gitter channel](https://gitter.im/assetgraph/assetgraph?at=5c669897253c2b5ea317dc75) it would be nice to allow customizing the precache service worker generated by `sw-precache`. 

`loadSwPrecacheConfig` was copypasted and tweaked from `bundleWebpack`. It is very similar but not the same because the webpack config is mandatory and this one is not. We might still want to extract it into a reusable function, but it's not trivial.

One thing I don't like about `configPath` that it has to be absolute or relative to the main binary (eg `buildProduction` or `mocha` in tests). The former is not convenient, the latter is not very useful. Btw the same is true for `bundleWebpack`'s `configPath` which is why I followed the pattern. 

As a user of any cli tool I would expect an explicit relative config path to be relative to the current working directory or the project root (the latter actually happens when using the canonical file name fallback). Wdyt?


papandreou:
Thanks a lot for doing this, looks solid to me!

> As a user of any cli tool I would expect an explicit relative config path to be relative to the current working directory or the project root (the latter actually happens when using the canonical file name fallback). Wdyt?

Yeah, you're right, the `require.main.require` trick doesn't really work as expected when `buildProduction` is involved, only when assetgraph is used programmatically from the project that's being built.

I think I've only used `buildProduction --webpackconfig` with absolute paths for that reason. We should try to fix up both I think. Maybe that's just a matter of throwing in `pathModule.resolve(process.cwd(), configPath)` since an absolute `configPath` will remain unchanged.

papandreou:
I did a bit of testing, and it seems like the `staticFileGlobs` config option isn't honored. It must be because we pass in the assets using the `dynamicUrlToDependencies` option: https://github.com/assetgraph/assetgraph/blob/403e0a0025f1141131f3f79162969db8ffda041c/lib/transforms/addPrecacheServiceWorker.js#L146

For now I've made the transform throw if that option is passed, so it's at least not silently ignored: 403e0a0025f1141131f3f79162969db8ffda041c

I don't know if it's a bug or if we should be checking each asset against the `staticFileGlobs` if given. That would be a bit fiddly.

papandreou:
> Maybe that's just a matter of throwing in `pathModule.resolve(process.cwd(), configPath)` since an absolute `configPath` will remain unchanged.

Did that in 1a471712902f51db7a584445a3627c2009dc33d2 :crossed_fingers:

PR #984, opened 2019-02-24
getTextByFontProperties: Don't require the AssetGraph instance

papandreou:
Makes it possible to use it in a browser.


Munter:
I'm having a hard time understanding the diff from a phone, but I trust our tests of this general area quite a lot. Let's merge

PR #987, opened 2019-03-03
autoprefixer: Add errorIfNotFound option

papandreou:
Defaults to true to match previous behavior

### DIFF
--- a/lib/transforms/autoprefixer.js
+++ b/lib/transforms/autoprefixer.js
@@ -19,9 +19,9 @@ module.exports = (
       } catch (e) {
         if (errorIfNotFound) {
           e.message =
-            `autoprefixer transform: Found ${
-              cssAssets.length
-            } css asset(s) while --browsers option is active, ` +
+            `autoprefixer transform: Found ${cssAssets.length} css asset${
+              cssAssets.length === 1 ? '' : 's'
+            }, ` +
             `but no autoprefixer module is available. Please use npm to install autoprefixer in your project so ` +
             `the autoprefixer transform can require it.\n${e.message}`;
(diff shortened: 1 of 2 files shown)


lib/transforms/autoprefixer.js
```diff
@@ -17,13 +17,17 @@ module.exports = (
...
-            cssAssets.length
-          } css asset(s) while --browsers option is active, ` +
-          `but no autoprefixer module is available. Please use npm to install autoprefixer in your project so ` +
-          `the autoprefixer transform can require it.\n${e.message}`;
-        throw e;
+        if (errorIfNotFound) {
+          e.message =
+            `autoprefixer transform: Found ${
+              cssAssets.length
+            } css asset(s) while --browsers option is active, ` +
```
Munter: It's weird that we reference `--browsers` here, in a transform in a project where the concept doesn't even exist.

lib/transforms/autoprefixer.js
```diff
@@ -17,13 +17,17 @@ module.exports = (
...
+            `but no autoprefixer module is available. Please use npm to install autoprefixer in your project so ` +
+            `the autoprefixer transform can require it.\n${e.message}`;
+          throw e;
```
Munter: Isn't the convention in other transforms where we do optional work that we emit a warning instead?

lib/transforms/autoprefixer.js
papandreou: Good point, fixed in 300c99b162ac9363df54f671890259b31a7195f4

papandreou:
Release in 5.8.0.

PR #999, opened 2019-03-31
subsetFonts: Async load the original @font-face declarations for self-hosted fonts

papandreou:
Fixes Munter/subfont#51


Munter:

PR #1000, opened 2019-03-31
 subsetFonts: Consistently issue root-relative urls instead of relative ones

papandreou:
It's a better default now that we support "mount root".


Munter:

PR #1003, opened 2019-04-07
JavaScriptDynamicImport relation (BLOCKED)

Munter:
This PR adds a new `JavaScriptDynamicImport`-relation to [dynamic import](https://github.com/tc39/proposal-dynamic-import)

Since none of the underlying tools we use have native support for this I had to attempt to look a bit into the future.

- [x] ~~Added [acorn-dynamic-import](https://www.npmjs.com/package/acorn-dynamic-import) to be able to parse the text in the first place.~~ *No longer needed, acorn supports dynamic import without a plugin now*
- [x] Awaiting estree/estree#198
- [x] Awaiting estools/escodegen#395
- [x] Awaiting estools/estraverse#100
- [x] Awaiting new acorn release
- [x] Awaiting new estraverse release

Even though the current implementation doesn't really work for real-world use cases, it actually improves the error message for the Assetgraph consumer. So it might actually be worth merging it with disabled tests to get that feature


lib/assets/JavaScript.js
```diff
@@ -537,6 +537,20 @@ class JavaScript extends Text {
...
+            node.type === 'CallExpression' &&
+            node.callee.type === 'Import' &&
+            node.arguments[0] &&
+            node.arguments[0].type === 'Literal' &&
+            typeof node.arguments[0].value === 'string'
+          ) {
+            if (isImportUrl(node.arguments[0].value)) {
+              outgoingRelations.push({
+                type: 'JavaScriptDynamicImport',
+                node: node,
```
papandreou:
```suggestion
                node,
```

test/relations/JavaScriptDynamicImport.js
```diff
@@ -0,0 +1,85 @@
...
+        __dirname,
+        '../../testdata/relations/JavaScriptDynamicImport/relative/'
+      )
+    });
+    const [javaScript] = await assetGraph.loadAssets('index.js');
+    await assetGraph.populate();
+    expect(javaScript.text, 'to contain', `import('./main.js');`);
+    assetGraph.findAssets({ fileName: 'main.js' })[0].url = `${
+      assetGraph.root
+    }/static/foobar.js`;
```
papandreou: `assetGraph.root` always has a trailing slash.

```suggestion
    }static/foobar.js`;
```

papandreou:
Let's see if this attempt gets accepted: https://github.com/estools/escodegen/pull/395

papandreou:
I updated the PR to align with estree/estree#198

Munter:
Awesome. One down...

papandreou:
Yay, finally! :tada:

papandreou:
Released in 5.12.0.

Munter:
Awesome! Thank you for pushing the other projects to get it into the entire ecosystem :D

Munter:
I guess this means we need to to a bit of tweaking in the bundleRelations transform to treat this `JavaScriptDynamicImport` as a new bundle discriminator

papandreou:
That shouldn't be necessary, the bundle discriminators only govern which scripts can be bundled on the basis of the incoming `<script>` and `<style>`s etc. It should be fine to bundle a JavaScript asset containing an `import(...)` with another script?

PR #1004, opened 2019-04-07
Check for missing glyphs in available formats

JesseFarebro:
If you want to subset a font to only `woff2`, assetgraph will throw an error due to hardcoding the missing glyph check for only the `woff` subset. This just checks the first subset as all glyphs will be in both subsets.

I also added a quick test case for the issue.


papandreou:
Excellent, thanks a lot for the contribution!

PR #1005, opened 2019-04-07
Keep inline assets when minifying css

JesseFarebro:
I discovered inlining subsetted fonts is currently not working. After debugging the issue it seems that when you minify a CSS asset it sets the `parseTree` with a reference to the current `parseTree`. When that happens it unloads all outgoing relations which are inline. When we get back from unloading the current `parseTree` we set the new tree equal to one where the inline assets have already been unloaded (it was a reference to the current tree). This causes the assets to not be included in the final output.

After sifting through the code I'm inclined to believe inline assets should be preserved across minification as they don't get repopulated. To this end, I think cloning the tree and operating on a copy of the tree fixes the issue.

I wrote a couple of test cases that would currently fail under master.


papandreou:
Thanks a lot for indentifying and looking into this problem!

It turns out that the bug was buried a bit deeper, it actually seems to be a regression in 4.0.0 where we changed the way unloaded assets are handled. Fixed that in 39f1e6cec96828ebe208da46ff80659dc5342f40 and got your test to pass without the cloning.

Released in 5.8.4

JesseFarebro:
@papandreou dang, I wasn't tipped off to anything in `AssetGraph` during my debugging session. Thanks for the quick responses and fixing the underlying issue.

papandreou:
Yeah, I can see why postcss/cssnano looked like the most likely culprit. It's pretty subtle that the base64 assets were actually removed from the AST as a side effect of the `this.unload()` in the `parseTree` setter because the involved relations happen to implement the `detach` method.

JesseFarebro:
I did get that far but never looked into how `removeAsset` was actually resolving the removal of inline assets. Thanks again! 👍 It's working great!

Issue #299, opened 2015-10-22
Make relations pluggable


papandreou:
Right now all relations are wired into the `findOutgoingRelationsInParseTree` method in the asset, which has become almost unmaintainable in some cases and makes it hard to allow plugins to define relations.

Inspiration: https://github.com/estools/esquery

papandreou:
Fixed on master now.

PR #1013, opened 2019-05-02
Add support of data-src and data-srcset for image element

vineetit1991:
1) Check for src attribute of image, if not exist then look for data-src.
2) Check for srcset attribute of image, if not exist then look for data-srcset.

### DIFF
--- a/lib/assets/Html.js
+++ b/lib/assets/Html.js
@@ -667,8 +667,8 @@ class Html extends Text {
             }
           }
         } else if (nodeName === 'img') {
-          const srcAttributeValue = node.getAttribute('src');
-          const srcSetAttributeValue = node.getAttribute('srcset');
+          const srcAttributeValue = node.getAttribute('src') || node.getAttribute('data-src');
+          const srcSetAttributeValue = node.getAttribute('srcset') || node.getAttribute('data-srcset');
           if (srcAttributeValue) {
             addOutgoingRelation({
               type: 'HtmlImage',
--- a/lib/relations/HtmlImage.js
+++ b/lib/relations/HtmlImage.js
@@ -2,11 +2,18 @@ const HtmlRelation = require('./HtmlRelation');
 
 class HtmlImage extends HtmlRelation {
   get href() {
-    return this.node.getAttribute('src');
+    return this.node.getAttribute('src') || this.node.getAttribute('data-src');
   }
 
   set href(href) {
-    this.node.setAttribute('src', href);
+    if (this.node.hasAttribute('src')) {
+      this.node.setAttribute('src', href);
+    } else if (this.node.hasAttribute('data-src')) {
+      this.node.setAttribute('data-src', href);
+    } else {
+      this.node.setAttribute('src', href);
+    }
+
   }
 
   get decoding() {
--- a/lib/relations/HtmlImageSrcSet.js
+++ b/lib/relations/HtmlImageSrcSet.js
@@ -3,7 +3,13 @@ const HtmlRelation = require('./HtmlRelation');
 class HtmlImageSrcSet extends HtmlRelation {
   inline() {
     super.inline();
-    this.node.setAttribute('srcset', this.to.text);
+    if (this.node.hasAttribute("srcset")) {
+      this.node.setAttribute('srcset', this.to.text);
+    } else if (this.node.hasAttribute("data-srcset")) {
+      this.node.setAttribute('data-srcset', this.to.text);
+    } else {
+      this.node.setAttribute('srcset', this.to.text);
+    }
     this.from.markDirty();
     return this;
   }


papandreou:
Interesting, thanks for submitting this. I assume that `data-src` and `data-srcset` aren't standardized? Could you write a few lines about how they're being used?

vineetit1991:
Sure,

Problem:- Load Images only when it is in view

Suppose we have carousel and it has so many images so we should not want to load all images at a time.So we can use some custom attribute (I am using data-src and data-srcset) where I will put image url and when the image comes in view, will convert data-src to src and data-srcset to srcset so that browser will download the particular image.

vineetit1991:
@papandreou  could you please let me know your thoughts on this ?

papandreou:
For a while we've been aiming to only support relation types that have basis in web standards, although we'll occasionally consider de facto standards or homegrown stuff like `'/path/to/file.png'.toString('url')`.

In this case I don't think this use of `` and `` are generally applicable enough to include in assetgraph core. It would be a great use case for a plugin via [pluggable relations](https://github.com/assetgraph/assetgraph/issues/299). Since that hasn't materialized yet, may I suggest that you use the following hack in the mean time:

```js
const AssetGraph = require('assetgraph');

const htmlProto = AssetGraph.Html.prototype;
const orig = htmlProto.findOutgoingRelationsInParseTree;
htmlProto.findOutgoingRelationsInParseTree = function() {
  const relations = orig.call(this);
  for (const node of Array.from(this.parseTree.querySelectorAll('img'))) {
    const dataSrcAttributeValue = node.getAttribute('data-src');
    const dataSrcSetAttributeValue = node.getAttribute('data-srcset');
    if (dataSrcAttributeValue) {
      relations.push({
        type: 'HtmlImage',
        href: dataSrcAttributeValue,
        node
      });
    }
    if (dataSrcSetAttributeValue) {
      relations.push({
        type: 'HtmlImageSrcSet',
        to: {
          type: 'SrcSet',
          text: dataSrcSetAttributeValue
        },
        node
      });
    }
  }
  return relations;
};
```

vineetit1991:
ok Thanks

vineetit1991:
@papandreou  With the above code we can get the value but still library would not be able to set value of data-src and data-srcset back. To do that I feel we need add some relations based on attribute like below. 
```
   var HtmlDataSrcSetAttribute = (function () {
        function HtmlDataSrcSetAttribute(config) {
            assetGraph.HtmlRelation.call(this, config);
        }

        util.inherits(HtmlDataSrcSetAttribute, assetGraph.HtmlRelation);

        extendWithGettersAndSetters(HtmlDataSrcSetAttribute.prototype, {
            inline: function () {
                assetGraph.Relation.prototype.inline.call(this);
                this.node.setAttribute('data-srcset', this.to.text);
        
                this.from.markDirty();
                return this;
            },

            attach: function (asset, position, adjacentRelation) {
                this.node = asset.parseTree.createElement('img');
                this.attachNodeBeforeOrAfter(position, adjacentRelation);
                return HtmlRelation.prototype.attach.call(this, asset, position, adjacentRelation);
            }
        });
        return HtmlDataSrcSetAttribute.bind(assetGraph);
    })();

    var HtmlDataSrcAttribute = (function () {
        function HtmlDataSrcAttribute(config) {
            assetGraph.HtmlRelation.call(this, config);
        }

        util.inherits(HtmlDataSrcAttribute, assetGraph.HtmlRelation);

        extendWithGettersAndSetters(HtmlDataSrcAttribute.prototype, {
            get href() {
                return this.node.getAttribute('data-src');
            },
        
            set href(href) {
                this.node.setAttribute('data-src', href);
            },

            attach: function (asset, position, adjacentRelation) {
                this.node = asset.parseTree.createElement('img');
                this.attachNodeBeforeOrAfter(position, adjacentRelation);
                return HtmlRelation.prototype.attach.call(this, asset, position, adjacentRelation);
            }
        });
        return HtmlDataSrcAttribute.bind(assetGraph);
    })();
```

Please let me know if there is any other way to do this. We are using version assetgraph 2.17.4

papandreou:
Hmm, right, you'll need something like that if the relations have to be updateable. I added an `attributeName` property to the `HtmlImage` relation and backported it to v2 now. If you upgrade to 2.17.5 you should be able to:

```
const AssetGraph = require('assetgraph');

const htmlProto = AssetGraph.Html.prototype;
const orig = htmlProto.findOutgoingRelationsInParseTree;
htmlProto.findOutgoingRelationsInParseTree = function() {
  const relations = orig.call(this);
  for (const node of Array.from(this.parseTree.querySelectorAll('img'))) {
    const dataSrcAttributeValue = node.getAttribute('data-src');
    const dataSrcSetAttributeValue = node.getAttribute('data-srcset');
    if (dataSrcAttributeValue) {
      relations.push({
        type: 'HtmlImage',
        href: dataSrcAttributeValue,
        attributeName: 'data-src',
        node
      });
    }
    if (dataSrcSetAttributeValue) {
      relations.push({
        type: 'HtmlImageSrcSet',
        to: {
          type: 'SrcSet',
          text: dataSrcSetAttributeValue
        },
        attributeName: 'data-srcset',
        node
      });
    }
  }
  return relations;
};
```

vineetit1991:
Thanks @papandreou  for the quick response.

PR #1042, opened 2019-07-24
Resolve canonical URL's with no trailing slash as the assetgraph root

Munter:
…. Relates to https://github.com/Munter/hyperlink/issues/158

Previously a canonical URL without a trailing slash would not be resolved to the local file system, causing the graph to detour around live sites.

I'm thinking that we should probably also have some sort of resolution so it hits `/index.html` somehow. Currently `/index.html` and `/` are two different assets if I read this correctly

### DIFF
--- a/lib/AssetGraph.js
+++ b/lib/AssetGraph.js
@@ -245,6 +245,12 @@ class AssetGraph extends EventEmitter {
               this.root
             );
           }
+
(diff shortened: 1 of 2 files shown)


Munter:
Seems safe enough, and definitely an improvement over before. I'll merge to unblock the hyperlink stuff I'm working on

Munter:

PR #1047, opened 2019-09-01
Pluggable relations for JavaScript

papandreou:
Implements https://github.com/assetgraph/assetgraph/issues/299 for JavaScript

TODO/questions:

- [ ] Is there a better static method name than `handler`?
- [x] ~~Consider turning `handler` and all `findOutgoingRelationsInParseTree` into generators.~~ *Might try to do a separate experiment on this*


papandreou:
@Munter, this should be ready for a review now. I think the main question is whether to keep the `selector`/`handler` division here, or if we should dumb it down and adopt a `getRelationsFromElement` pattern. I think we have a better shot at optimizing the `selector`/`handler` thing with a custom-built function. I thought the `esdispatch` module would do that, but it didn't seem to :)

PR #1049, opened 2019-09-06
Remove Bluebird

Munter:
These days we shouldn't really need Bluebird anymore.

- [x] Replace bluebird with native promises where possible
- [x] Use `util.promisify` to promisify in stead of bluebird
- [x] Replace bluebird usage for concurrency. Maybe `p-] limit`?
- [x] Replace bluebird usage for `.finally` support in https://github.com/assetgraph/assetgraph/blob/master/lib/util/determineFileType.js#L12
- [x] Delete bluebird from package.json

I renamed the remaining uses of bluebird from `Promise` to `Bluebird` to make it easier to spot the leftover usages.

The `.finally` usage in https://github.com/assetgraph/assetgraph/blob/master/lib/util/determineFileType.js#L12 is a homegrown concurrency limiter. Could that whole thing also be replaced with `p-limit` or another out of the box concurrency control for native promises?

### DIFF
--- a/lib/transforms/addDataVersionAttributeToHtmlElement.js
+++ b/lib/transforms/addDataVersionAttributeToHtmlElement.js
@@ -1,21 +1,19 @@
 const childProcess = require('child_process');
 const { promisify } = require('util');
+const exec = promisify(childProcess.exec);
 
 module.exports = (queryObj, version) => {
   return async function addDataVersionAttributeToHtmlElement(assetGraph) {
     if (typeof version === 'undefined') {
       // Try to derive a version tag from git:
       try {
-        const rawVersion = await promisify(cb =>
-          childProcess.exec(
-            'git describe --long --tags --always --dirty',
-            {
-              encoding: 'utf-8'
-            },
-            cb
-          )
-        )();
-        version = rawVersion.replace(/^[\s\n\r]+|[\s\r\n]+$/g, '');
+        const { stdout } = await exec(
+          'git describe --long --tags --always --dirty',
+          {
+            encoding: 'utf-8'
+          }
+        );
+        version = stdout.replace(/^[\s\n\r]+|[\s\r\n]+$/g, '');
       } catch (e) {}
     }
     if (version) {
--- a/lib/transforms/addDataVersionAttributeToHtmlElement.js
+++ b/lib/transforms/addDataVersionAttributeToHtmlElement.js
@@ -1,16 +1,20 @@
 const childProcess = require('child_process');
+const { promisify } = require('util');
 
 module.exports = (queryObj, version) => {
   return async function addDataVersionAttributeToHtmlElement(assetGraph) {
     if (typeof version === 'undefined') {
       // Try to derive a version tag from git:
       try {
-        const rawVersion = childProcess.execSync(
-          'git describe --long --tags --always --dirty',
-          {
-            encoding: 'utf-8'
-          }
-        );
+        const rawVersion = await promisify(cb =>
+          childProcess.exec(
+            'git describe --long --tags --always --dirty',
+            {
+              encoding: 'utf-8'
+            },
+            cb
+          )
+        )();
         version = rawVersion.replace(/^[\s\n\r]+|[\s\r\n]+$/g, '');
       } catch (e) {}
     }
--- a/lib/transforms/addDataVersionAttributeToHtmlElement.js
+++ b/lib/transforms/addDataVersionAttributeToHtmlElement.js
@@ -1,20 +1,17 @@
 const childProcess = require('child_process');
-const Promise = require('bluebird');
 
 module.exports = (queryObj, version) => {
   return async function addDataVersionAttributeToHtmlElement(assetGraph) {
     if (typeof version === 'undefined') {
       // Try to derive a version tag from git:
       try {
-        const [stdout] = await Promise.fromNode(
-          cb =>
-            childProcess.exec(
-              'git describe --long --tags --always --dirty',
-              cb
-            ),
-          { multiArgs: true }
+        const rawVersion = childProcess.execSync(
+          'git describe --long --tags --always --dirty',
+          {
+            encoding: 'utf-8'
+          }
         );
-        version = stdout.replace(/^[\s\n\r]+|[\s\r\n]+$/g, '');
+        version = rawVersion.replace(/^[\s\n\r]+|[\s\r\n]+$/g, '');
       } catch (e) {}
     }
     if (version) {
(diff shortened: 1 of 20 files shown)


lib/transforms/addDataVersionAttributeToHtmlElement.js
```diff
@@ -1,20 +1,17 @@
...
       // Try to derive a version tag from git:
       try {
-        const [stdout] = await Promise.fromNode(
-          cb =>
-            childProcess.exec(
-              'git describe --long --tags --always --dirty',
-              cb
-            ),
-          { multiArgs: true }
+        const rawVersion = childProcess.execSync(
```
papandreou: Ouch, could we avoid `execSync`? Maybe something like:

```diff
diff --git a/lib/transforms/addDataVersionAttributeToHtmlElement.js b/lib/transforms/addDataVersionAttributeToHtmlElement.js
index 3debad62..bf958461 100644
--- a/lib/transforms/addDataVersionAttributeToHtmlElement.js
+++ b/lib/transforms/addDataVersionAttributeToHtmlElement.js
@@ -1,16 +1,20 @@
 const childProcess = require('child_process');
+const { promisify } = require('util');
 
 module.exports = (queryObj, version) => {
   return async function addDataVersionAttributeToHtmlElement(assetGraph) {
     if (typeof version === 'undefined') {
       // Try to derive a version tag from git:
       try {
-        const rawVersion = childProcess.execSync(
-          'git describe --long --tags --always --dirty',
-          {
-            encoding: 'utf-8'
-          }
-        );
+        const rawVersion = await promisify(cb =>
+          childProcess.exec(
+            'git describe --long --tags --always --dirty',
+            {
+              encoding: 'utf-8'
+            },
+            cb
+          )
+        )();
         version = rawVersion.replace(/^[\s\n\r]+|[\s\r\n]+$/g, '');
       } catch (e) {}
     }
```

lib/transforms/addPrecacheServiceWorker.js
```diff
@@ -214,7 +214,7 @@ module.exports = (
     } else {
       batches = entryPointHtmlAssets.map(asset => [asset]);
     }
-    await Promise.map(batches, generateAndAttachPrecacheServiceWorkerAsync, {
+    await Bluebird.map(batches, generateAndAttachPrecacheServiceWorkerAsync, {
```
papandreou: https://github.com/sindresorhus/p-map is a drop-in replacement.

lib/transforms/drawGraph.js
```diff
@@ -139,19 +138,19 @@ module.exports = targetFileName => {
     ]);
     dotProcess.stdin.end(dotSrc);
     dotProcess.stderr.on('data', chunk => console.warn(`dot STDERR: ${chunk}`));
-    await Promise.fromNode(cb => {
+    await new Promise(resolve => {
       dotProcess.on('error', error => {
         if (error.code === 'ENOENT') {
-          cb(new Error('"dot" not found. Please install graphviz.'));
+          throw new Error('"dot" not found. Please install graphviz.');
```
papandreou: These `throw`s will turn into unhandled exceptions. They should reject the promise instead.

lib/transforms/gzip.js
```diff
@@ -9,25 +10,23 @@ module.exports = queryObj => {
...
-    await Promise.map(
+    await Bluebird.map(
       gzipAssets,
       async asset => {
         // http://webmasters.stackexchange.com/questions/31750/what-is-recommended-minimum-object-size-for-gzip-performance-benefits
         if (asset.type !== 'SourceMap' && asset.rawSrc.length > 860) {
-          const gzippedRawSrc = await Promise.fromNode(cb =>
-            compress.gzip(asset.rawSrc, cb)
-          );
+          const gzippedRawSrc = await compress.gzip(asset.rawSrc);
```
papandreou: Hmm, does this work? It seems weird that `compress.gzip` would return a promise when only the main export of `node-zopfli` or `gzip` has been promisified?

We don't seem to have any tests covering this transform.

lib/transforms/loadAssets.js
```diff
@@ -10,7 +11,7 @@ module.exports = function(...args) {
   const flattenedArguments = _.flatten(args);
   return async function loadAssets(assetGraph) {
     return _.flatten(
-      await Promise.map(
+      await Bluebird.map(
```
papandreou: `p-map` should work fine here, too.

lib/transforms/gzip.js
Munter: Good catch. Fixed in aa133bbb

lib/transforms/drawGraph.js
Munter: Fixed in dcd95269

lib/transforms/addDataVersionAttributeToHtmlElement.js
Munter: Fixed in 43a1ae24

Why doesn't this work?

```js
const rawVersion = await promisify(childProcess.exec)(
  'git describe --long --tags --always --dirty',
  {
    encoding: 'utf-8'
  }
);
```

papandreou: Hmm, yeah, that's more elegant than involving a callback. It does work if you destructure the fulfillment value:

```js
        const { stdout } = await promisify(childProcess.exec)(
          'git describe --long --tags --always --dirty',
          {
            encoding: 'utf-8'
          }
        );
        version = stdout.replace(/^[\s\n\r]+|[\s\r\n]+$/g, '');
```

lib/transforms/addPrecacheServiceWorker.js
Munter: Done in 4fd02081

lib/transforms/loadAssets.js
Munter: Done in 4fd0208

lib/transforms/addDataVersionAttributeToHtmlElement.js
Munter: Ah! I'll make that change.

Munter: Done in b810eab7

lib/transforms/gzip.js
papandreou: Great! I added a gzip test on master (https://github.com/assetgraph/assetgraph/commit/eb007153b08ec91149bd684ff0b0a2dbb307f559), and it also passes on this branch :muscle:

papandreou:
@Munter, I tried to replace that hand-rolled concurrency limiter in `determineFileType` with `p-limit` + added a test in 2643cd2.

Munter:
Yay! Great to slim things down a bit. Then we just need https://github.com/One-com/teepee/pull/8 merged so we can get bluebird completely out of the non-development

PR #1051, opened 2019-09-09
Pluggable relations for Html, Xml, Svg, Rss, Atom, and MsApplicationConfig

papandreou:
Essentially the same pattern as #1047

Discussion items:

- [x] The `stopTraversal` option is only used by `HtmlSvgIsland`. Can we come up with a better solution? I have been playing around with using selectors with namespaces, but couldn't make it fit the use case.
- [x] Currently the selector-based syntax only allows matching elements, so it's not possible to convert the relations that live in HTML comment nodes (`HtmlConditionalComment`, `HtmlKnockoutContainerless`, `HtmlEdgeSideIncludeSafeComment`). Is that worth fixing now? My best bet is to register a selector for `*` and poke through the child nodes.

Small breaking change:

* `XmlHtmlInlineFragment` has been split into `AtomHtmlInlineFragment` and `RssHtmlInlineFragment`.


Munter:
I was thinking that the relations current `selector` getter would actually be more of a test method that takes a node and tests for a match itself. Then it's up to the individual test method to choose between `node.matches(selector)` or a more low-level `nodeType` comparison like we had in `Html` before

papandreou:
Hmm, yeah, that would be more flexible, and then we wouldn't even need to split it into two static methods.

Since only elements have a `matches` method, it would result in a bit of boilerplate, though:

```js
static getRelationsFromNode(node) {
  if (node.nodeType === node.ELEMENT_NODE && node.matches('a[href]')) {
    return {
      type: 'HtmlAnchor',
      // ...
    };
  }
}
```

Munter:
```js
static getRelationsFromNode(node) {
  if (node.matches?.('a[href]')) {
    return {
      type: 'HtmlAnchor',
      // ...
    };
  }
}
```

The future is not far away :D https://github.com/tc39/proposal-optional-chaining#syntax

papandreou:
Hmm, yeah, that's not too bad :)

Then it should be fine to consult `getRelationsFromNode` for all nodes, including text, comments and processing instructions.

What about the pluggable JavaScript relations, then? Is it too opinionated to bake in `esquery`? We don't really have the problems with nodes that escape selectors there.

Munter:
Looks great

papandreou:
The last remaining question is how we can get to move the handling of conditional comments out of `Html#findOutgoingRelationsInParseTree`. Right now we attach a *conditionalComments* property to all other relations in order to use it as a "bundle discriminator": https://github.com/assetgraph/assetgraph/blob/5c4889d555e9a011080f88c21ea88e048c679990/lib/assets/Html.js#L362-L367

And a little below that we do the housekeeping: https://github.com/assetgraph/assetgraph/blob/d7e9dc97aaf84d78ffd958f048e74eaf7a361e8f/lib/assets/Html.js#L373-L395

As I see it we can either:

1) Keep it like this so that conditional comments enjoy special support in core.
2) Add extra complexity to the interface between `Html` and its relation classes so that relations can keep track of a stack during the traversal *and* somehow intercept other relations so a property can be attached.
3) Update the two transforms that use the `conditionalComments` property to do a bit more work so we don't need it.

I think I vote for 3)

papandreou:
Implemented option 3 now, and also introduced a handy `Html#_visitAllNodesInDom` generator :)

Munter:
Beautiful!

papandreou:
Changed #1052 to match the changes. It's tempting to have a common superclass for XML and HTML to avoid the duplication.

PR #1063, opened 2019-11-16
Add hack to recover from missing xml namespace declarations

Munter:
Following up on discussion starting at https://gitter.im/assetgraph/assetgraph?at=5dcfe4ca92a84f79fe791bd6

an inline Svg asset inside html will throw a parse error if using a namespace that is not declared within the inline svg. This happens even though the namespace is a known one in browsers, which handle the case just fine.

Example:

```html
<svg><use xlink:href="icons.svg#foo"></use></svg>
```

This very hacky addition injects a fake namespace declaration for the missing namespace in the element where it is needed. It modifies the internal asset text:

```html
<svg><use xlink:href="icons.svg#foo" xmlns:xlink="http://fakenamespace.nonstandard"></use></svg>
```

This is obviously not properly tested yet, and since it is a major hack I'm not quite happy with the solution, so lets discuss our options


Munter:
@papandreou I think this approach is the right one. The tests are failing right now because I'm injecting the list of namespaces unconditionally. I think I can improve on this to detect used namespaces instead and inject them if needed

Munter:
Switched over to auto detect which namespaces are used and inject all of the undeclared ones. This passes all tests, including my failure case in hyperlink. Missing test coverage for this functionality. Will look at that later

papandreou:
I made a `v6` branch in case you want to backport it to get it released.

PR #1065, opened 2019-12-09
Support <base href=...> when resolving urls in Html (and inline assets)

papandreou:
Came up in https://github.com/Munter/subfont/issues/71

Does not implement the weirdo rule that `<base href=...>` only applies to elements following it. My estimate is that there are zero use cases for that.

Maybe we can implement some kind of warning if there are multiple `<base href=...>`  in the document, and maybe even if it comes after a relation, but it seems outrageous to implement the full spec.


lib/assets/Asset.js
```diff
@@ -571,6 +571,17 @@ class Asset extends EventEmitter {
     return null;
   }
 
+  get baseUrl() {
+    if (this.isInline) {
+      const nonInlineAncestor = this.nonInlineAncestor;
+      if (nonInlineAncestor) {
+        return this.nonInlineAncestor.baseUrl;
```
Munter: This is a pretty nice place to put it. It's certainly more intuitive to just access `asset.baseUrl` instead of remembering that in some cases it can be inline and then having to specifically look for `nonInlineAncestor`

Munter:
I'll test this linked into munter/subfont#71 when I get home, to get the real-life case tested as well

lib/assets/Asset.js
papandreou: Yes, I agree!

Although we do need to keep `Relation#baseUrl` as well because `sourceRoot` in source maps only affects the `SourceMapSource` relation: https://github.com/assetgraph/assetgraph/blob/3f1b3d2602f69e120fa43ddb4d495e53da658e6f/lib/relations/SourceMap/SourceMapSource.js#L12-L22

papandreou:
I guess we need to backport it to v6 to try it out with subfont. Luckily it applies cleanly: https://github.com/assetgraph/assetgraph/pull/new/tech/backportBaseHrefToV6 :relieved:

papandreou:
Made a PR for that: https://github.com/assetgraph/assetgraph/pull/1066

Munter:
Looks like everything checks out after running the sapper case too. Lets merge!

papandreou:
Ah, great! :muscle:

PR #1074, opened 2020-01-05
Support reading the encoding of XML documents from <?xml ... encoding=...?>

papandreou:


lib/assets/Xml.js
```diff
@@ -1,7 +1,57 @@
 const JSDOM = require('jsdom').JSDOM;
 const Text = require('./Text');
 
+function extractEncodingFromText(text) {
+  const matchEncoding = text.match(/^<\?xml[^>]* encoding="([^"]*)"[^>]*\?>/i);
```
Munter: I'm not really good at reading EBNF. Are newlines allowed here? https://www.w3.org/TR/xml/#xmldoc

papandreou: Ah, you're right, the _S_ (white space) token also allows tab, carriage return, and newline: https://www.w3.org/TR/xml/#NT-S

Fixed in b256338

Issue #1081, opened 2020-02-26
calling loadAssets with wildcard loads directories


palindrom615:
## as-is

I tried to execute `subfont` with argument `**/*.html` which is the default value of `subfont`.

And this is the error message:
```
page-data/404.html - Text._getTextFromRawSrc(): No _rawSrc property found: page-data/404.html
```

The error occurs because `page-data/404.html` is not a file but a directory.

this is where the error occurs:

https://github.com/Munter/subfont/blob/457c7f0e4cef0a8c1bd8f816c23ace64c9987424/lib/subfont.js#L132
```
  await assetGraph.loadAssets(inputUrls);
```

## to-do

I hesitated to report the issue because I am not sure if a directory is kind of an asset or not... 

If a directory is not an asset we want, why doesn't it just check and filter if it is a file or not?

Munter:
This is certainly a bug, and thanks for reporting it. I'm a bit unsure if a safeguard check against directories should be a part of Assetgraph core, or it it is the individual tools responsibility to filter out inputs that wont work. Any thoughts on this @papandreou ?

papandreou:
Hmm, yeah, I'm not really sure. It feels a bit weird to handle this at the assetgraph level if the tool really did tell it to add the directory to the graph. But maybe we can try to be more resilient and just leave the asset unloaded in that case.

Munter:
Looks like the problem is that we are using https://www.npmjs.com/package/glob and it also returns directories as matches. I'll configure it to not do that

Munter:
The fix is released in assegraph@6.0.6

Thank you for reporting

PR #1091, opened 2020-05-17
Avoid directories in glob results in loadAssets and addAssets. Fixes …

Munter:
…#1081


Munter:
Merging. I'll pick it to `v6` as well

Munter:
Released in v6.0.6

PR #1122, opened 2020-10-31
Update fetch expression to only match direct descendants.

kevincox:
Previously it would match cases like `f("hi", fetch(url))` and consider the target `hi`.

### DIFF
--- a/lib/relations/JavaScript/JavaScriptFetch.js
+++ b/lib/relations/JavaScript/JavaScriptFetch.js
@@ -3,7 +3,7 @@ const Relation = require('../Relation');
 
 class JavaScriptFetch extends Relation {
   static get selector() {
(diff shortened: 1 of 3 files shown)


papandreou (approved):
Thanks! :rocket:

papandreou:
I'm guessing this doesn't need to be backported to 6.x?

kevincox:
🤷 I haven't looked into anything but the latest version.

papandreou:
Does that mean that you're running directly off master? It's currently a few hundred commits ahead of the latest released version, 6.2.1, and it doesn't look like this bug exists in 6.2.1: https://github.com/assetgraph/assetgraph/compare/v6.2.1..master#diff-34995cc9684380629526a9e450a27102e25023763dc58da81fa80437a81a8274L536-L566

kevincox:
I didn't realize they were that far apart. I guess I'm switching back and forth whenever I do `npm link` to debug something. I guess I should probably pin to the release version then 😅

papandreou:
Yeah, I've been working towards making the whole thing more pluggable for a while, but there's still a lot more to do :sweat_smile:

Issue #1123, opened 2020-10-31
Problem with baseUrl and JavaScriptFetch, JavaScriptServiceWorkerRegistration...


kevincox:
Right now assets have a `baseUrl` which defaults to their `url`, or the `url` of the `baseUrl` of their nearest non-inline ancestor. However this is a flawed concept for JavaScript as it has a much more complex situation than that. Notably `import` is resolved relative to `baseUrl` however `fetch`, `navigator.serviceWorker.register` and similar are relative to the URL of the including document. Currently the code treats these as the same, which results in unfortunate results such as `navigator.serviceWorker.register('file:///home/kevincox/p/kevincox.ca/build/service-worker.js')`. 

However this problem is basically impossible unless assetgraph knows all possible references to the JavaScript file. There are cases where we can probably assume this such as when using `assetgraph-hashfiles` however there are also cases where this can not be assumed, such as if you are providing a JavaScript library that you expect users (and possibly your website as well) to access off of a well-known URL.

I encountered this problem with `serviceWorker.register` but I believe the same issue exists with `fetch`.

I think the best solution here is to not modify the referenced URL when the URL of the asset *containing* the reference is changed. There are a couple of cases where we can do something "clever" by handling the case where the only user of the script is moved, however this action at a distance is probably more confusing than it is helpful.

kevincox:
Since I knew that all of my callers were coming from the domain of the target the following hack seems to be a viable workaround. A better solution would probably be remembering the original URL of the relation and always using that as the `baseUrl`.

```js
function fixJavaScriptDocumentRelativeRelations(Class) {
	Object.defineProperties(Class.prototype, {
		hrefType: {
			get() { return "rootRelative" },
			set() {},
		},
		baseUrl: {
			get() { return this.assetgraph ? this.assetgraph.root : this.url },
		},
	})
}
fixJavaScriptDocumentRelativeRelations(AssetGraph.JavaScriptFetch);
fixJavaScriptDocumentRelativeRelations(AssetGraph.JavaScriptServiceWorkerRegistration);
```

papandreou:
Yeah, we used to have a really complex implementation that tried to get this right, but it still didn't always work due to the ambiguities you're pointng out. Would it be an option for you to use root-relative urls when `fetch`ing static assets and registering service workers?

kevincox:
That is what my work around is doing. It seems like the "ideal" solution is fairly easy to define, but hard to implement:

[...]

PR #1125, opened 2020-11-01
Add test demonstrating issue with javascript baseUri confusion.

kevincox:
See https://github.com/assetgraph/assetgraph/issues/1123

### DIFF
--- a/test/relations/JavaScript/JavaScriptFetch.js
+++ b/test/relations/JavaScript/JavaScriptFetch.js
@@ -174,4 +174,24 @@ describe('relations/JavaScriptFetch', function () {
       'JavaScriptFetch.attach(): Not implemented'
     );
   });
+
+  it('should not change when script moves.', async function () {
+    const assetGraph = new AssetGraph({
+      root: pathModule.resolve(
+        __dirname,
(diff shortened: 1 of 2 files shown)


papandreou:
Hmm, yeah, I guess it isn't really wrong that it turns into an absolute `file://` url, it's just kind of impractical in most use cases :laughing:.

It appears to be what [this hack](https://github.com/assetgraph/assetgraph/blob/898d0b136f2651986b086aea033a832f87c62e94/lib/relations/JavaScript/JavaScriptStaticUrl.js#L24-L34) tries to fix for `JavaScriptStaticUrl` relations. I've [expanded it to cover the other JavaScript relations where the url is interpreted relative to the HTML asset](https://github.com/assetgraph/assetgraph/commit/39c543f33f96072f5fc04102806acbecd7f45c7a) now. That makes it come out as `fetch('/b.js');` -- I hope that does the trick?

kevincox:
That should fix my instance of the problem. However it does seem still like a hack more than a solution.

papandreou:
Yeah, agreed. It's just a bit hard to figure out how to handle it in a way that's both generic and correct.

PR #1127, opened 2020-11-14
Stop modelling fetch(...) as a relation

papandreou:
https://github.com/Munter/hyperlink/pull/189
https://github.com/assetgraph/assetgraph-builder/blob/547cb460a3a1060a29ac3c588751d5cb049f030c/lib/transforms/buildProduction.js#L115


papandreou:
Okay, I'll make an executive decision here :)

Hopefully by the time we make a new major release with this change, it'll be possible to make a plugin that adds `fetch` support if someone actually wants that.

PR #1180, opened 2021-07-11
Add asset XmlSitemap and relation XmlSitemapUrl to enable population …

Munter:
…of sitemaps

### DIFF
--- a/lib/assets/Asset.js
+++ b/lib/assets/Asset.js
@@ -515,6 +515,13 @@ class Asset extends EventEmitter {
           }
         }
       }
+
+      if (type === 'Xml' && typeof this._text === 'string') {
+        if (this._text.includes('<urlset') && this._text.includes('xmlns="http://www.sitemaps.org/schemas/sitemap/')) {
+          type = 'XmlSitemap';
+        }
+      }
+
       this._tryUpgrade(type);
 
       this.emit('load', this);
--- a/lib/assets/Asset.js
+++ b/lib/assets/Asset.js
@@ -516,12 +516,6 @@ class Asset extends EventEmitter {
         }
       }
 
-      if (type === 'Xml' && typeof this._text === 'string') {
-        if (this._text.includes('<urlset') && this._text.includes('xmlns="http://www.sitemaps.org/schemas/sitemap/')) {
-          type = 'XmlSitemap';
-        }
-      }
-
       this._tryUpgrade(type);
 
       this.emit('load', this);
--- a/lib/assets/Xml.js
+++ b/lib/assets/Xml.js
@@ -2,6 +2,9 @@ const JSDOM = require('jsdom').JSDOM;
 const Text = require('./Text');
 
 class Xml extends Text {
+  /**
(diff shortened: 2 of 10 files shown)


lib/assets/Asset.js
```diff
@@ -515,6 +515,13 @@ class Asset extends EventEmitter {
           }
         }
       }
+
+      if (type === 'Xml' && typeof this._text === 'string') {
```
papandreou, line 519: Eeeek, having this in the base class will make the plugin-ification effort more difficult (although we could of course expand the api surface area to support it). Is there really no other way to infer that an asset is a sitemap?

What are some use cases for getting these relations populated? Maybe we could consider a transform-based approach.

Munter: We should definitely consider a new lifecycle hook for pluginification here.

There is only one way I know of where we can infer the type from a relation, and that would be if we also implement support for `robots.txt`, where you can do `Sitemap: https://example.com/sitemap.xml`. But I feel like it would be a perfectly valid use case to use the sitemap as an entry point for assetgraph, in which case we would only have the file content to go on for inferring the type.

papandreou: What is the exact use case, though?

Munter: Give assetgraph a sitemap to start the crawl instead of `index.html`. In my case, the sitemap includes all landing pages, which are not linked to tom the normal navigation hierarchy, so they wouldn't be discoverable unless I point them out explicitly. And they are not on disk, so hard to discover

papandreou: What are you trying to do with the graph? Is it for hyperlink?

Munter: Right now I'm making some custom tools for Heap that check done extra details about links and redirects, plus that the canonical is set correctly for google indexing. I need a practical way to start the graph, and this at least follows an established specification instead of me writing a custom query for our headless CMS to start the graph.

And I do indeed intend to use sitemaps in hyperlink as well. It would be a great addition to the default assets it starts from, if there is a sitemap present

papandreou: Okay, makes sense! It's a bit of an invasive change as it stands. Would you consider an "detect and upgrade sitemaps" transform instead?

Munter: I could live with a transform for now, if the next major had a lifecycle/plugin hook that would let me move the upgrade login into the sitemap asset

papandreou: I can't make any specific promises about the next major version, the exact scope is still TBD.

Munter:
@papandreou I switched the sitemap asset upgrade to a transform as per our previous discussion. It makes the usage slightly unintuitive, but I'm hoping we can hide that in the higher-level tools like hyperlink

papandreou:
Great! Could you turn it into a "plugin" then so it doesn't have to reside in core?

Munter:
Closing, since this has been pulled into a plugin: https://github.com/assetgraph/assetgraph-plugin-sitemap

PR #1182, opened 2021-07-15
Ensure _text is always available on JavaScript assets.

alexjeffburke:
As of assetgraph 7.1.0 assignment to the .text property keeps the
original source code available as _text. This is immensely useful
but has also opened up an unfortunate inconsistency because _text
is not populated from the rawSrc when such assets are instantiated.

In order to avoid adding special cases to callers try to make _text
uniformly available assetgraph is able to decode the raw buffer.

### DIFF
--- a/lib/assets/JavaScript.js
+++ b/lib/assets/JavaScript.js
@@ -12,6 +12,12 @@ class JavaScript extends Text {
   init(config) {
     super.init(config);
     this.serializationOptions = this.serializationOptions || {};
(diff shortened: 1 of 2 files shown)


papandreou:
I'd rather not have to do this extra work in the constructor, `_text` and `_rawSrc` are just supposed to be cached representations.

papandreou:
Can you take a step back and explain the use case?

Issue #1226, opened 2022-03-13
Dependabot complaint due to perfectionist dependency


danielfdickinson:
hyperlink (due to assetgraph) is giving me dependabot warnings (CVE-2021-23382) due to ``hyperlink@5.0.4 requires postcss@^5.0.8 via a transitive dependency on perfectionist@2.4.0``

https://github.com/ben-eb/perfectionist was last updated five years ago, and has been archived by the author.

I presume it will be a 'bit' of a pain to eliminate this dependency, so I'm wondering if there is an ETA on a version of assetgraph that eliminates the problem? I thought I saw somewhere you had a new major version in the works; am I mistaken (and if not, any ETA?).

I really love your hyperlink package which depends on assetgraph, so I am hoping this is something that is in the works, though that is obviously entirely dependent on your time, energy, interest, etc.

Cheers, and thanks for the great work!

papandreou:
I have not looked into it yet, do you know if there's a natural successor to `perfectionist` in the `postcss` ecosystem?

danielfdickinson:
@papandreou The short answer is no. The long answer is that the only thing I could find was a few comments on the postcss repo, starting here: <https://github.com/postcss/postcss/issues/387#issuecomment-726102750>.

papandreou:
For now I forked `perfectionist`, made it postcss 8-compatible, and released assetgraph 7.4.0 with that. Should be an in-range update since `hyperlink@5.0.4` depends on `assetgraph@^7.3.1`. I hope that helps.

danielfdickinson:
Thank you, again, @papandreou! This appears to be working well, and since tonight I am doing some updates, if there are more complicated cases that have broken (unlikely), then I should be able to let you know soon.

danielfdickinson:
Sorry, spoke too soon. The fix helped, but now there is:

```text
hyperlink@5.0.4 requires postcss@^8.0.9 via a transitive dependency on css-declaration-sorter@6.1.4
```

which comes through https://github.com/assetgraph/assetgraph/blob/4cffe642ab1d05d78ac63970ccbe6652749bce48/package.json#L28

and

```text
hyperlink@5.0.4 requires postcss@^5.2.4 via a transitive dependency on postcss-scss@0.3.1
```

which is in the new forked package (probably copied from original ``perfectionist``).

papandreou:
I tried updating `perfectionist-papandreou` to update `postcss-scss` to `^4.0.3`. Seems to work fine.

[...]

PR #1227, opened 2022-03-19
Replace perfectionist with perfectionist-papandreou (postcss 8 compatible)

papandreou:
Context: https://github.com/assetgraph/assetgraph/issues/1226


papandreou:
Released in 7.4.0

PR #1231, opened 2022-03-21
Refresh CHANGELOG.md with preversion command

akx:
I noticed the CHANGELOG file hadn't been regenerated with any versions since 6.0.

Maybe a more automated process for this would be nice...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,158 @@
-### v6.0.0 (2019-08-17)
+### v7.5.0 (2022-03-20)
+
+- [#1228](https://github.com/assetgraph/assetgraph/pull/1228) Update more postcss-related packages ([Andreas Lind](mailto:andreas.lind@workday.com))
+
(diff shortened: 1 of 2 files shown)


papandreou:
Hmm, interesting, there's already a `preversion` hook in place for that: https://github.com/assetgraph/assetgraph/blob/e3e25848e0311533b1202c621941a91b153e508f/package.json#L111 🤔

papandreou:
Ah, wait, it's because I'm releasing from the `v7` branch: https://github.com/assetgraph/assetgraph/blob/v7/CHANGELOG.md

It has drifted quite a bit from `master`, so that's why I haven't been merging changes (including `CHANGELOG.md`) back to `master`. Will try to remember that going forward.

akx:
I see. Is there a reason why v7 isn't developed in `master`?

papandreou:
> I see. Is there a reason why v7 isn't developed in `master`?

Yeah, long story. Essentially `master` is being used to develop v8.

PR #1241, opened 2022-06-02
[review] Add support for Safari mask-icon.

bramstein:
Safari has support for SVG favicons through the `mask-icon` relation type. The SVG file is treated as a single-color mask, and automatically adapts to light & dark mode. You can read more about it in the [Apple mask-icon documentation](https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariWebContent/pinnedTabs/pinnedTabs.html).

Assetgraph doesn't currently support this type of relation so processing a site using `mask-icon` results in a broken icon in Safari. This pull request adds support for the `mask-icon` relationship so that the SVG file is correctly included in the assetgraph output.

### DIFF
--- a/lib/relations/Html/HtmlShortcutIcon.js
+++ b/lib/relations/Html/HtmlShortcutIcon.js
@@ -8,7 +8,7 @@ class HtmlShortcutIcon extends HtmlRelation {
     if (
       node.nodeType === node.ELEMENT_NODE &&
       node.matches(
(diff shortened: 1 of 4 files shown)


papandreou (approved):
LGTM 🚀 

If you need this to work now, we'll need to backport it to the `v7` branch and do a new release.

bramstein:
If it's not too much trouble, that would be great.

bramstein:
Awesome, thank you!

